PR #3, opened 2017-02-04
Updates README.md: neatly aligns "Rank" union type

mariusschulz:
TypeScript allows an optional leading `|` in union types, which you can use to align the string literal types in the `Rank` more neatly.


pelotom:
Thanks!

PR #5, opened 2017-02-24
implement Dictionary value runtime type checking

bbenezech:
To make things simple, I modelled the type behaviour on TypeScript’s:

- JS objects that are not an array are a string dictionary
- JS objects with keys that can be casted to number are also a number dictionary
- JS arrays are number dictionaries

By default dictionaries are string based, but can optionally be number based with the second parameter.

The PR is feature complete as far as I can see.


src/index.ts
```diff
@@ -228,17 +228,29 @@ export function Tuple(...runtypes: Runtype<any>[]) {
...
+      throw new ValidationError(`Expected a defined non-null value but was ${typeof xs}`)
+    for (const key of Object.keys(xs)) {
+      // Object.keys is always a string array
+      if (keyType === 'number') {
+        if (isNaN(+key))
+          throw new ValidationError(`Expected dictionary key to be a number but was string`)
+      }
+      if (keyType === 'string') {
+        if (''+(+key) === key)
+          throw new ValidationError(`Expected dictionary key to be a string but was number`)
```
pelotom: I'm not sure this particular check is valid. This means that we disallow string dictionaries with any keys that happen to be a string representation of a number. TypeScript treats this as valid, for instance:

```ts
const foo: { [_: string]: string } = { x: 'blah', 3: 'hi' }
```

and I'd prefer to stay as close to the intended semantics of the static type system as possible.

src/index.ts
```diff
@@ -247,7 +259,8 @@ export function Dictionary<V>() {
  */
 export function Record<O>(runtypes: {[K in keyof O]: Runtype<O[K]> }): Runtype<O> {
   return runtype(x => {
-    Dictionary().check(x)
+    if (x === null || x === undefined)
+      throw new ValidationError(`Expected a defined non-null value but was ${typeof x}`)
```
pelotom: How about:

```ts
Dictionary(Always).check(x)
```

bbenezech: It won't work. Dictionary do not accept strings, arrays, number, etc.
Record({}) does.

pelotom: Oh right, of course. In that case, let's use

```ts
Record({}).check(xs)
```

in the definition of `Dictionary`.

pelotom:
Thanks for the contribution, looks great!

PR #6, opened 2017-02-24
Dictionary doc

bbenezech:
Thanks for merging PR.
Dictionary doc additions, I tried to stay on topic ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ type Fleet = {
   [shipId: number]: Ship
 }
 
-type SpaceObject = Asteroid | Planet | Ship | Fleet | StarSystem
+type SpaceObject = Asteroid | Planet | Ship | StarSystem | Fleet
 ```
 
 If the objects which are supposed to have these shapes are loaded from some external source, perhaps a JSON file, we need to
--- a/README.md
+++ b/README.md
@@ -54,7 +54,15 @@ type Ship = {
   crew: CrewMember[]
 }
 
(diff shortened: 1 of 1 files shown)


pelotom:
Hm, I feel like this makes the example overly complicated. The example doesn't need to be exhaustive in its use of the available types, it just needs to get the basic idea across.

What the documentation could really use right now is a section for each type, describing it and giving an example usage along with the static type that it corresponds to.

bbenezech:
I think I would keep examples somewhat feature complete as long as they are the only source of documentation.

Improving the documentation with a feature index may be premature effort-wise:
- the API is small
- the example is focused and clear
- TS makes parameters and details self-discoverable once you found the entry points 

I had no issue at all to get started thanks to the example, but if Dictionary had existed then, I would have wasted time not finding it in the doc and having to look it up in the source.

But it's your call and you should probably trust your judgement: DX-wise, this project is simply awesome: API, tests and Readme are absolutely **fabulous** ❤️ 

Thanks a lot for your efforts open sourcing this code.
Let me know if you need some help maintaining it in the future.

pelotom:
I've added a dictionary to the expanded [space-game](https://github.com/pelotom/runtypes/blob/master/src/examples/space-game.ts) example in the `examples` directory. In general, if you want to add examples, big or small, to that directory, I would be more than happy to accept them, because they are not just passive documentation; they get type checked!

Thanks for your kind words and above all for your willingness to contribute!

PR #11, opened 2017-03-30
Add key to error message of value's runtype.

DanielBrookRoberge:
This PR adds additional documentation to the type checking errors in `Record` and `Optional` types. If one of the key checks fails, the `Record` will add `In key ${key}:` to the beginning of the string, indicating which of the keys has the problem.


pelotom:
Nice, thanks!

PR #16, opened 2017-06-03
Add optional correction fn and add withConstraint to Contract

typeetfunc:
Hi, I am working on https://github.com/typeetfunc/runtypes-generate. This project allows generating data by Runtype type with JSVerify(https://github.com/jsverify/jsverify). Because of it, I want to introduce 2 features:
1) Pass optional `correction` and `rightInverse` fn to `Constraint` constructor. This fns uses in generating data by `Constraint` type(https://github.com/typeetfunc/runtypes-generate/blob/master/src/index.ts#L12)
2) Add `.withConstraint` to Contract type and add argsType and returnType to an internal struct of `Contract` type. This is necessary cause I want to be able to generate data for function's `argsType` and `returnType`.
See example in the test: https://github.com/typeetfunc/runtypes-generate/blob/master/src/index.spec.ts


typeetfunc:
Related: https://github.com/gcanti/tcomb/issues/263

pelotom:
This sounds cool! I'll take a look asap, probably after the weekend.

typeetfunc:
@pelotom Can we make `Contract` regular Runtype instance?
I want to be able to pass `Contract` to `Contract`. Example:
```js
const FindContract = Contract(Contract(String, Boolean), Array(String)).enforce(filterFn)
```
In the Contract` implementation:
```js
return returnType.check(f(...args.map((item, i) => argTypes[i].tag === 'contract' ? argsTypes[i](item) : item)))
```
What do you think?

src/types/constraint.ts
```diff
@@ -1,12 +1,19 @@
...
   tag: 'constraint'
-  underlying: A
+  underlying: A,
+  constraint: ConstraintCheck<A>,
+  correction?: ConstraintCorrection<A>,
+  rightInverse?: ConstraintRightInverse<A>
 }
 
-export function Constraint<A extends Rt>(underlying: A, constraint: (x: A) => boolean | string) {
+export function Constraint<A extends Rt>(underlying: A, constraint: ConstraintCheck<A>, correction?: ConstraintCorrection<A>, rightInverse?: ConstraintRightInverse<A>) {
```
pelotom: This looks like a bug carried over from the previous code, but shouldn't this be

```ts
export function Constraint<A extends Rt>(underlying: A, constraint: ConstraintCheck<Static<A>>, correction?: ConstraintCorrection<Static<A>>, rightInverse?: ConstraintRightInverse<Static<A>>) {
```

pelotom:
Re. making `Constraint` a runtype, assume you're suggesting something like

```ts
Constraint<A, B>(...): Runtype<(x: A) => B>`
```

is that right? The thing that feels awkward about this is, what does `check`/`validate`/`guard` do for this runtype? It's undecidable whether an arbitrary function meets some specification, so our only option is to say that any function passes validation, only to possibly throw runtime errors later when someone tries to use it. That's why `Contract` exists as a separate concept from runtypes, whereas `Function` is a simple unparameterized runtype that only promises to validate that a value is some kind of function (`Runtype<(...args: any[]) => any>`), which is the best it can do.

It may be a reasonable compromise to make it a no-op runtype, but it feels a little bit weird.

src/contract.ts
```diff
@@ -162,13 +175,24 @@ export function Contract(...runtypes: Runtype<any>[]) {
...
-  return {
+  let argsTuple = Tuple.apply(null, argTypes)
+
+  const result = {
+    tag: 'contract',
+    argTypes: argsTuple,
+    returnType,
+    withConstraint: (...args: any[]) => {
+      argsTuple = argsTuple.withConstraint(...args)
+      result.argTypes = argsTuple
```
pelotom: Mutating `result` is not ideal

pelotom: (also reassigning `argsTuple`)

PR #17, opened 2017-06-12
feat(types/custom.ts): Add Custom(or Brand) type

typeetfunc:
Rethinking approach from https://github.com/pelotom/runtypes/pull/16
PR added `Custom`(maybe `Brand` or `Predicate`?) type. See example: https://github.com/typeetfunc/runtypes-generate/blob/master/src/custom.spec.ts#L23
Main idea: give an opportunity to create user-defined types with any predicate. But unlike `Constraint`, `Custom` save the name of predicate and args, and we can write a generator for this type in test(https://github.com/typeetfunc/runtypes-generate/blob/master/src/custom.spec.ts#L35).
What do you think about this approach? Also, It might be a good idea for runtypes create `runtypes` scope in npm and `runtypes` org in GitHub for collecting Custom types(`Contains`, `Range`, `Email`, `Integer` etc) with generators.


src/types/custom.ts
```diff
@@ -0,0 +1,20 @@
+import { Runtype, Rt, Static, create, validationError } from '../runtype'
+import { String } from './string'
+
+export interface Custom<A extends Rt, T extends string, K> extends Runtype<Static<A>> {
+  tag: T,
+  underlying: A,
+  args: K
+}
+
+export function Custom<A extends Rt, T extends string, K>(underlying: A, constraint: (x: Static<A>, args: K) => boolean | string, tag: T, args: K) {
```
pelotom: Why does the `constraint` argument need to be

```ts
constraint: (x: Static<A>, args: K) => boolean | string
```

instead of just

```ts
constraint: (x: Static<A>) => boolean | string
```

You're already defining `constraint` within the scope of `args`, so you don't need to pass it in, do you?

src/types/custom.ts
```diff
@@ -0,0 +1,20 @@
+import { Runtype, Rt, Static, create, validationError } from '../runtype'
+import { String } from './string'
+
+export interface Custom<A extends Rt, T extends string, K> extends Runtype<Static<A>> {
+  tag: T,
```
pelotom: `tag` should just be `'custom'`, because it's meant to be used when doing case analysis across the closed set of built-in runtypes. I like the idea of having a custom type as an escape hatch out of the closed world, but doesn't the `args: K` field already give you the power to stuff in whatever information you want? For example, you could have `args: { tag: 'my-custom-type', x: number, y: boolean}`.

typeetfunc: >doesn't the args: K field already give you the power to stuff in whatever information you want?
Yes, you are right.
>tag should just be 'custom', because it's meant to be used when doing case analysis across the closed set of built-in runtypes
Hmm, i thought it was more properly. Why do you need a predefined and closed set of built-in runtypes? I'm not trying to argue. I'm just interested in your opinion :)

typeetfunc:
>Couldn't we just augment Constraint to have optional additional args and let it fulfill the role of both
Ok, I will do so. Thanks for suggestion!:)

src/types/custom.ts
typeetfunc: I thought it was more DRY. See example in test: https://github.com/typeetfunc/runtypes/blob/feature/custom-type/src/index.spec.ts#L60

src/types/custom.ts
pelotom: So that you can write functions against a `Reflect` and know that all the possible alternatives (what the values of `tag` might be) have been handled. For example, look at the definition of [`show`](https://github.com/typeetfunc/runtypes/blob/feature/custom-type/src/show.ts). It defines a switch statement across all possible run types and it doesn't need a `default` case because it knows it's handled everything. If you deleted any of those cases or if you added another alternative to `Reflect` it would be a static type error (assuming your compiler settings are appropriately strict), which is nice because it alerts you to potential bugs.

src/types/custom.ts
pelotom: You can write that as

```ts
CustomArray: (args => Custom(Array(Number), (x, args) => x.length > args.min, 'Length', args))({min:3})
```

if you really need to write it inline like that without repeating yourself.

typeetfunc:
@pelotom fixed :)

typeetfunc:
@pelotom fixed

lib/runtype.d.ts
```diff
@@ -32,7 +32,9 @@ export interface Runtype<A> {
      * for some reason. May also return a string which indicates an
      * error and provides a descriptive message.
      */
-    withConstraint(constraint: (x: A) => boolean | string): Constraint<this>;
+    withConstraint<K extends {
+        tag: string;
+    }>(constraint: ConstraintCheck<this>, args?: K): Constraint<this, K>;
```
pelotom: Get rid of the constraint on `K` here?

typeetfunc: @pelotom fixed

Issue #18, opened 2017-06-23
No way to skip optional field check


vgammieri:
Hi,

I tried the following:
```typescript
const OptionalString = String.Or(Undefined);
type OptionalString = Static<typeof OptionalString>;


const IOrganization = Record({
    id: NonEmptyString,
    name: NonEmptyString,
    description: OptionalString,
    link: OptionalString,
});
type IOrganization = Static<typeof IOrganization>; 
```
if I call
```typescript
const o: IOrganization = {
      id: "a",
      name: "the name",
};
IOrganization.check(o);
```
The `check` function throws an error! There is a way to avoid the throwing if an optional field is missing in the argument object?

typeetfunc:
https://github.com/pelotom/runtypes/issues/15
```js
const A = Record({
  T: String,
  N: Number
}).And(Partial({
  Z: Boolean
}));
```
Just use this pattern(`Z` will optional)

vgammieri:
I'm trying by following this approach:
```typescript
export function Record<O extends { [_: string]: Rt }>(fields: O) {
  return create<Record<O>>(x => {
    if (x === null || x === undefined)
      throw validationError(`Expected a defined non-null value but was ${typeof x}`)

    // tslint:disable-next-line:forin
    for (const key in fields) {
      if (hasKey(key, x)) {
        try {
          fields[key].check(x[key])
        } catch ({ message }) {
          throw validationError(`In key ${key}: ${message}`)
        }
      } else {
        const reflect = fields[key] as any;
        if (reflect.tag === "union") {
          const alternatives: any[] = reflect.alternatives;
          const foundUndefined = alternatives.filter((a) => a.tag === "literal" && a.value === undefined);
          if (foundUndefined.length === 0) {
            throw validationError(`Missing property '${key}'`)
          }
        }

      }
    }

    return x as O
  }, { tag: 'record', fields })
}
```

pelotom:
Please see [this comment](https://github.com/pelotom/runtypes/issues/15#issuecomment-307601297) for why your proposal is inconsistent with the static semantics of TypeScript. As @typeetfunc said, the correct way to do this is by intersecting a `Record` and a `Partial`.

Issue #15, opened 2017-05-31
"Better" support for nullable properties in Record?


spacepope:
Hi,

at first i didn't know, how to declare nullable properties in a Record and if it is possible at all.
The only way i found so far is:
```
const A = Record({
  T: String,
  N: Number
}).And(Partial({
  Z: Boolean
}));
```
which results in a type:
```
type AType = {
    T: string;
    N: number;
} & {
    Z?: boolean;
}
```

Wouldn't it be more easy and intuitive if one could do:
```
const A = Record({
  T: String,
  N: Number,
  Z: Nullable(Boolean) // or Boolean.AsNullable()
  /*
       or even better as this is the normal typescript syntax?:
       Z?: Boolean
  */
});
```

Something like that shouldn't be that hard, or do i miss something?

Thanks and regards..

pelotom:
Yes, the `Record.And(Partial)` is the current best practice for achieving this. I like the `Nullable` idea, unfortunately I'm pretty sure TypeScript's type system is insufficiently powerful to express it. 

Think about what the type signature would have to be for the `Record` type constructor in order to infer the right type as an output: you basically want to tell the type system to go through the value types and pick out the ones that were constructed with a special `Nullable` constructor, and for _just those_ the value type of the record should be optional.

I would be delighted to learn that there's a way to do this sort of thing, but AFAIK it's not possible :(

spacepope:
Hmm, ok i see..
Would the following be a good or a bad idea:

There is already an Undefined and a Void runtype.
What if in Record code (Line 19) is only checked for key existance if the value is not of type Undefined or Void?
So one could define an optional record property with  `NullableProperty: Number.Or(Void)`..
What do you think?

pelotom:
Allowing keys to be missing if their corresponding value types are `null` or `undefined` is inconsistent with the static type semantics of typescript:

```ts
const foo: { x: null | undefined } = {}
//    ^^^
// Type '{}' is not assignable to type '{ x: null | undefined; }'.
//  Property 'x' is missing in type '{}'.
```

spacepope:
To correct myself: i was rather speaking of optional than nullable properties..
But as it seems (and besides your comment) one cannot selectively pick properties from an underlying type filtered to the value type. And if it would be possible, it probably results in the union of non optional and optional parameters too, so i can live with the way it works..

PR #19, opened 2017-06-23
Check if a missing field in a record is declared as optional

vgammieri:


pelotom:
Thanks for your PR, but I'm afraid I can't accept it; see response to #15 and #18.

Issue #22, opened 2017-08-10
Possible to make a RunType that uses instanceof?


breathe:
Sometimes I'm expecting an instance of a particular class.  Is there a way to make a constraint field based on instanceof check or other typeguard?

I want to express something like this ...

```
class PopulationDistribution { ... }

const Planet = Record({
  type: Literal('planet'),
  location: Vector,
  mass: Number,
  population: InstanceOf(PopulationDistribution),
  habitable: Boolean
})
```

pelotom:
I think it's certainly doable and I'd welcome a PR if you wanted to take a crack at it :)

breathe:
Ok I'll give it a shot!

breathe:
Closed by #23

PR #23, opened 2017-08-14
Attempting to add an InstanceOf runtype

breathe:
Towards resolving: Issue #22 

Not mergeable yet -- but here's a first attempt to add a InstanceOf runtype.  The tests I added pass but I'm not entirely sure I'm taking the right approach to this...  I'm new to typescript still and this level of metaprogramming is a little confusing -- though this project's code is so clean that it seems a great place to learn from.

For reasons I didn't figure out yet the test cases I added in show.spec.ts produce type-errors (the compiler error is why the tests don't pass in CI).  If you ignore the compiler error -- the tests do pass when run ...

If you get a chance, let me know what you think!

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -23,8 +23,9 @@ import {
   Contract,
   Reflect,
   InstanceOf,
-  IInstanceOfFunction
 } from './index'
+
+import { Constructor } from "./types/InstanceOf"
 
 const boolTuple = Tuple(Boolean, Boolean, Boolean)
 const record1 = Record({ Boolean, Number })
@@ -291,7 +292,7 @@ describe('reflection', () => {
   | Intersect2<Reflect, Reflect>
   | Function
   | Constraint<Reflect, any>
-  | InstanceOf<IInstanceOfFunction>
+  | InstanceOf<Constructor>
 ): Reflect => {
   const check = <A>(X: Runtype<A>): A => X.check({})
   switch (X.tag) {
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -22,6 +22,7 @@ import {
   Constraint,
   Contract,
   Reflect,
+  InstanceOf
 } from './index'
 
 const boolTuple = Tuple(Boolean, Boolean, Boolean)
@@ -30,6 +31,9 @@ const union1 = Union(Literal(3), String, boolTuple, record1)
 
 type Person = { name: string, likes: Person[] }
 const Person: Runtype<Person> = Lazy(() => Record({ name: String, likes: Array(Person) }))
+
+class SomeClass {}
+class SomeOtherClass {}
 
 const runtypes = {
   Always,
@@ -60,7 +64,10 @@ const runtypes = {
   CustomArrayWithMessage: Array(Number).withConstraint(x => x.length > 3 || `Length array is not greater 3`, {tag: 'lenght', min:3}),
   Dictionary: Dictionary(String),
   NumberDictionary: Dictionary(String, 'number'),
-  DictionaryOfArrays: Dictionary(Array(Boolean))
+  DictionaryOfArrays: Dictionary(Array(Boolean)),
+  InstanceOfSomeClass: InstanceOf(SomeClass),
+  InstanceOfSomeOtherClass: InstanceOf(SomeOtherClass),
+  DictionaryOfArraysOfSomeClass: Dictionary(Array(InstanceOf(SomeClass)))
 }
 
 type RuntypeName = keyof typeof runtypes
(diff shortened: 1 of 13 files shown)


src/reflect.ts
```diff
@@ -20,3 +20,4 @@ export type Reflect =
   | { tag: 'intersect'; intersectees: Reflect[] } & Runtype<always>
   | { tag: 'function' } & Runtype<(...args: any[]) => any>
   | { tag: 'constraint'; underlying: Reflect; constraint: ConstraintCheck<Reflect>; args?: any } & Runtype<always>
+  | { tag: 'instanceof', ctor: Reflect } & Runtype<any>
```
pelotom: Should be `ctor: Function` probably? The underlying constructor is not a runtype, so it won't have a `Reflect`.

src/index.spec.ts
```diff
@@ -275,6 +290,7 @@ describe('reflection', () => {
   | Intersect2<Reflect, Reflect>
   | Function
   | Constraint<Reflect, any>
+  | InstanceOf<Reflect>
```
pelotom:
```ts
InstanceOf<Function>
```

src/index.spec.ts
```diff
@@ -322,6 +338,10 @@ describe('reflection', () => {
       break
     case 'constraint':
       check<Static<typeof X.underlying>>(X)
+      break;
+    case 'instanceof':
+      check<Static<typeof X.ctor>>(X)
```
pelotom:
```ts
check<typeof X.ctor>(X)
```

breathe:
Thanks a lot for the feedback and sorry for the delayed response!

I incorporated those changes -- it still wasn't compiling ...  (compilation would fail in index.spec.ts):
```
src/index.spec.ts(347,3): error TS2322: Type 'Always | String | Never | Void | Boolean | Number | Function | Literal<string | number | boolean>...' is not assignable to type 'Reflect'.
  Type 'InstanceOf<Function>' is not assignable to type 'Reflect'.
    Type 'InstanceOf<Function>' is not assignable to type '{ tag: "instanceof"; ctor: Function; } & Runtype<any>'.
      Type 'InstanceOf<Function>' is not assignable to type '{ tag: "instanceof"; ctor: Function; }'.
        Types of property 'ctor' are incompatible.
          Type 'Function' is not assignable to type 'Function'. Two different types with this name exist, but they are unrelated.
            Property 'apply' is missing in type 'Function'.src/index.spec.ts(347,3): error TS2322: Type 'Always | String | Never | Void | Boolean | Number | Function | Literal<string | number | boolean>...' is not assignable to type 'Reflect'.
  Type 'InstanceOf<Function>' is not assignable to type 'Reflect'.
    Type 'InstanceOf<Function>' is not assignable to type '{ tag: "instanceof"; ctor: Function; } & Runtype<any>'.
      Type 'InstanceOf<Function>' is not assignable to type '{ tag: "instanceof"; ctor: Function; }'.
        Types of property 'ctor' are incompatible.
          Type 'Function' is not assignable to type 'Function'. Two different types with this name exist, but they are unrelated.
            Property 'apply' is missing in type 'Function'.
```

I believe the above happens due to the fact that the lexical symbol 'Function' is shadowed by the `Function` symbol import 

```import { ... Function ...} from "./index")```

I changed the code to use an interface IInstanceOfFunction (simply an interface that extends from Function).  I did this so that I could verify that issue cause.  It seems like maybe this is a good pattern to avoid this kind of ambiguity in the future?

Given that the Function runtype is public api -- maybe its a good idea to change the exported name of the runtype ...?  I could see errors around this kind of thing being ... --- really hard to track down in 3rd party code?

breathe:
Hi @pelotom -- I force-pushed a change to my repo at some point (after squashing) -- I believe that may have broken the review workflow on this PR ...  Any decision on whether you want to add this feature?

pelotom:
Hi @breathe, sorry for the delay, I'll try to get around to reviewing this again today.

pelotom:
> I believe the above happens due to the fact that the lexical symbol 'Function' is shadowed by the Function symbol import
> `import { ... Function ...} from "./index")`

Have you tried using a renaming import to sidestep this? `import { Function as FunRuntype } from './index'`

Another thing I commonly do is use `(...args: any[]) => any` in place of `Function`.

My philosophy so far has been to use pretty names even if they might conflict with the builtin types, allowing people to use renaming imports if they need to disambiguate.

breathe:
>Have you tried using a renaming import to sidestep this? import { Function as FunRuntype } from './index'

That would work - though I would obviously have to change other uses of Function in that file to FunRunType.  I thought that using the interface was more clear and would make for better error messages during development.  Do you want the implementation changed?

pelotom:
I guess I'm ok with it, but I'm not a big fan of the I-prefix for interfaces. What about calling the interface to just `Constructor`?

pelotom:
Oops, accidentally closed 😛

breathe:
@pelotom Just pushed the rename from IInstanceOfFunction to Constructor.  I also changed index.ts to export only InstanceOf but not the `Constructor` disambiguation interface -- I believe this interface doesn't need to be exported outside the library -- the name is just used internally to disambiguate vs the 'Function' runtype ...

pelotom:
Great, this looks good!

Issue #36, opened 2018-02-01
Suggestions around pattern matching


twop:
First of all, thank you for putting such an amazing library. It utilizes both typescript and javascript really well producing such a empowering and complete developer experience :)

I found you project while looking at the ways to pattern match in typescript. So my suggestions are biased.

Let's say we have this:

```typescript
const One = Record({ tag: Literal('one'), descOfOne: String });
const Two = Record({ tag: Literal('two'), descOfTwo: String });

type OneOrTwo = Static<typeof One> | Static<typeof Two>;
```

While matching I want to extract payload as well. Something like:

```typescript
const DescriptionOfOne = One.pipe(x=>x.descOfOne); 
//not Runtype anymore, can be used only for pattern matching

// OneOrTwo | null -> string 
const getDescription = matchCase<OneOrTwo | null>()
  .of(DescriptionOfOne, desc => desc) // acceptes extracted prop
  .of(Two, ({descOfTwo}) => descOfTwo) // normal types can be used just fine
  .of(Void, ()=> 'nothing');
```

DescriptionOfOne has different type now. Lets say Pattern<One, string>.

```typescript
const NonEmptyDesc = DescriptionOfOne.withContraint( desc => desc.lenght >0) 
// Error! Will not compile. Patterns can only be mapped over.

const OneDescLenght = DescriptionOfOne.map(decs => desc.lenght);
// totally valid though.
```

I think it makes sense to map them further to produce different pipelines.

But if we use Records extensively for matching we have perf issues :( "check" check all props recursively.

```typescript
const TaggedOne = Tag(One, 'tag'); //new Runtype that remembers which one is the tag
// disables deep tree matching and checks only tags. 
// useful for application types (like redux actions)
// It might make sense to introduce Tagged({shape}) and TaggedLiteral('one') for that purpose as well.
```

Also with typescript 2.8 we will have a proper Diff<T,U>. And if I understand it correctly it opens this
```typescript
const EmptyOne = One.withConstraint( one => one.descOfOne.length === 0);

//Possible from typescript 2.8 with a help of Diff
const getDescription = matchCase<OneOrTwo | null>()
  .of(EmptyOne, _ => 'empty one') // leaves the same type
  .guard(One, _ => "one") // removes One, so after we can guard only Two | Void
  .guard(Two, _ => "two") // leaved just Void to check
  .guard(Void, ()=> 'nothing'); // never after this point
```

[...]

Issue #8, opened 2017-03-05
Return multiple errors


OliverJAsh:
Would it be possible for `validate/check` to return all errors instead of just the first?

This would be useful for validating user input, giving the user all errors instead of just one.

pelotom:
That's an interesting idea, I think it would make sense as an optional flag passed to `check`.

OliverJAsh:
If the validator functions didn't throw, it might be easier to run all the validators and then aggregate the errors.

An optional flag to `validate/check` seems sensible to me.

bbenezech:
@OliverJAsh `check` doesn't throw.

pelotom:
@bbenezech `check` throws, `validate` doesn't.

It probably makes sense to let `check` just continue to throw the first error it encounters, and change `validate`'s `Result` to give you an array (or `Iterable`) of errors.

khmelevskii:
@pelotom are you going to implement it? I think it's really important to return all validation errors instead of first error

pelotom:
> @pelotom are you going to implement it? I think it's really important to return all validation errors instead of first error

I don’t have time to work on this myself, happy to review a PR though.

schicks:
I'd be happy to PR this. What should the behavior be on union types though? I would lean towards reporting the error for every variant of the union, but that means that fixing one error out of N could fix the other N-1.

pelotom:
I'd say when in doubt, WWTSD (what would typescript do)? Model it after how typescript reports static errors.

nthypes:
Do you accept a PR on this? This is a huge downside and avoid a lot of users to use **runtypes**.

scotttrinh:
For inspiration, here is how `io-ts` added returning all errors to their latest experimental API: https://github.com/gcanti/io-ts/commit/d68dbf13d2ebe87974f8cb220882329672dedff2

nthypes:
Well, if anyone still having this issue, I'm building a lib which solves this.

https://github.com/nthypes/super-ts

ivawzh:
I gave it a try. See WIP PR here 👉  https://github.com/ivawzh/runtypes/pull/1

The amount of work is much bigger than I thought. Before continuing further, I want to bring up some discussions.

There will be breaking changes with the current approach. E.g. `Failure` format.

[...]

PR #37, opened 2018-02-05
Incremental checking + parallel pattern matching

pelotom:
Switches the checking method of composite runtypes to be incremental, using generators. This allows pattern matching to be performed in parallel across cases, which solves the main performance concern raised in #36.

The parallel pattern matching algorithm is roughly as follows:
- in parallel, incrementally check each case
- if a case finds an error, remove it from further consideration
- if a case has finished checking (i.e. was successful) and is the topmost case still under consideration, use that case's handler
- if all but one case has been eliminated, immediately use that case's handler with no further checking. (_"When you have eliminated the impossible, whatever remains, however improbable, must be the truth."_)

A further optimization that could be considered is transforming the provided cases into mutually exclusive cases so that no duplicate work is performed in parallel.

Incremental checking also lays a natural foundation for #8, because the user can keep pulling out as many errors as they like from the runtype's generator.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -46,41 +46,41 @@ class SomeOtherClass {
 }
 
 const runtypes = {
-  Always,
-  Never,
-  Undefined,
-  Null,
-  Empty: Record({}),
-  Void,
-  Boolean,
-  true: Literal(true),
-  false: Literal(false),
-  Number,
-  3: Literal(3),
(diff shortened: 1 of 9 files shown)


pelotom:
General-purpose pattern matching has not proved to be useful enough in Runtypes to make it worth the effort to so radically complicate the type checking algorithm, closing.

PR #51, opened 2018-04-16
More descriptive errors

henrinormak:
Goal was to improve the legibility of error messages, with emphasis on nested complex types.

There are things that could be improved further, for example, in union types errors from all attempts should be captured, not just the general "Expected one of ..., but was ..." as that is often super difficult to use in a debugging context (it would be much easier to understand why none of the types suited, not just the fact that none of them did).

There's also likely some room to expand the error messaging in Partial type, but this would be improved automatically with changes/improvements to Union (as it uses Union under the hood).

Some examples of the new errors

```
import { Array, String, Number, Boolean, Record } from 'runtypes';
const Flags = Array(Boolean);
Flags.check([true, true, 's']);
// { Error: Expected boolean but was string, key: '[2]' }

const Page = Record({
  title: String
});

const Person = Record({
  name: String,
  age: Number,
  likes: Array(Page)
});
const People = Array(Person);
People.check([{ name: 'Jack', likes: [] }]);
// { Error: Expected age to be number, key: '[0].age' }

People.check([{ name: 'Jack', age: 10, likes: [{ name: undefined, likes: [] }] }]);
// { Error: Expected title to be string, key: '[0].likes.[0].title' }

const Pet = Record({
  name: String
});

const PersonOrPet = Union(
  Person,
  Pet
);

PersonOrPet.check({ foo: 'bar' });
// { Error: None of the alternatives { name: string; age: number; likes: { title: string; }[]; } | { name: string; } matched {"foo":"bar"} }

const PartialPerson = Partial({
  name: String,
  age: Union(Number, String),
  likes: Array(Page)
});

PartialPerson.check({ name: 'Jack', age: { value: 10 } });
// { Error: Expected (number | string) | undefined, got {"value":10} key: 'age' }
```

There is also a bit of discussion on performance that can be seen on https://github.com/6fold/runtypes/pull/1

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -311,7 +311,7 @@ describe('check errors', () => {
         name: String,
         age: Number,
       }),
-      'Expected number | undefined, but was object',
+      'Expected number, but was null',
       'age',
     );
   });
@@ -324,8 +324,8 @@ describe('check errors', () => {
         age: Number,
         likes: Array(Record({ title: String })),
       }),
-      'Expected { title: string; }[] | undefined, but was object',
-      'likes',
+      'Expected string, but was number',
+      'likes.[0].title',
     );
   });
 
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -69,11 +69,13 @@ const runtypes = {
   Person,
   MoreThanThree: Number.withConstraint(n => n > 3),
   MoreThanThreeWithMessage: Number.withConstraint(n => n > 3 || `${n} is not greater than 3`),
+  ArrayString: Array(String),
   ArrayNumber: Array(Number),
+  ArrayPerson: Array(Person),
   CustomArray: Array(Number).withConstraint(x => x.length > 3, { tag: 'lenght', min: 3 }),
   CustomArrayWithMessage: Array(Number).withConstraint(
     x => x.length > 3 || `Length array is not greater 3`,
-    { tag: 'lenght', min: 3 },
+    { tag: 'length', min: 3 },
   ),
   Dictionary: Dictionary(String),
   NumberDictionary: Dictionary(String, 'number'),
@@ -106,9 +108,14 @@ const testValues: { value: always; passes: RuntypeName[] }[] = [
   { value: { Boolean: true, foo: 'hello' }, passes: ['Partial'] },
   { value: { Boolean: true, foo: 5 }, passes: [] },
   { value: (x: number, y: string) => x + y.length, passes: ['Function'] },
+  { value: { name: undefined, likes: [] }, passes: [] },
+  { value: { name: 'Jimmy', likes: [{ name: undefined, likes: [] }] }, passes: [] },
   { value: { name: 'Jimmy', likes: [{ name: 'Peter', likes: [] }] }, passes: ['Person'] },
   { value: { a: '1', b: '2' }, passes: ['Dictionary'] },
-  { value: ['1', '2'], passes: ['NumberDictionary'] },
+  { value: ['1', '2'], passes: ['ArrayString', 'NumberDictionary'] },
+  { value: ['1', 2], passes: [] },
+  { value: [{ name: 'Jimmy', likes: [{ name: 'Peter', likes: [] }] }], passes: ['ArrayPerson'] },
+  { value: [{ name: null, likes: [] }], passes: [] },
   { value: { 1: '1', 2: '2' }, passes: ['Dictionary', 'NumberDictionary'] },
   { value: { a: [], b: [true, false] }, passes: ['DictionaryOfArrays'] },
   { value: new Foo(), passes: [] },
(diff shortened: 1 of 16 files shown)


henrinormak:
@pelotom - if the code coverage drop is critical, I can figure out to increase it (it mostly dropped due to some if-else branches not being covered in cases with nested error keys).

pelotom:
@henrinormak if you could restore the coverage that would be awesome 😄

src/index.spec.ts
```diff
@@ -179,6 +186,217 @@ describe('contracts', () => {
   });
 });
 
+describe('check errors', () => {
+  it('tuple type', () => {
+    try {
+      const Check = Tuple(Number, String, Boolean);
+      Check.check([false, '0', true]);
+      fail("tuple passed, eventhough should've failed");
```
pelotom: "even though"

src/index.spec.ts
```diff
@@ -179,6 +186,217 @@ describe('contracts', () => {
...
+    } catch (e) {
+      expect(e.message).toBe('Expected a number, but was boolean');
+      expect(e.key).toBe('[0]');
+    }
```
pelotom: This same structure is repeated for all the error message checks, how about extracting a function to eliminate the repetition and make clear what in particular is being tested in each case?

src/types/tuple.ts
```diff
@@ -223,10 +223,25 @@ export function Tuple<
...
+      try {
+        xs = Arr(Always).check(x);
+      } catch ({ key, message }) {
+        throw validationError(`Expected tuple to be an array: ${message}`, key);
+      }
+
       if (xs.length < components.length)
-        throw validationError(`Expected array of ${components.length} but was ${xs.length}`);
-      for (let i = 0; i < components.length; i++) components[i].check(xs[i]);
+        throw validationError(`Expected an array of ${components.length}, but was ${xs.length}`);
```
pelotom: an array of length*

src/types/union.ts
```diff
@@ -965,7 +966,9 @@ export function Union(...alternatives: Rt[]): any {
   return create(
     x => {
       for (const { guard } of alternatives) if (guard(x)) return x;
-      throw new Error('No alternatives were matched');
+
+      const a = create<any>(x as never, { tag: 'union', alternatives });
+      throw validationError(`Expected a ${show(a)}, but was ${typeof x} ${JSON.stringify(x)}`);
```
pelotom: Remove the "a" after "Expected"

pelotom: Using `JSON.stringify` is not safe because it can throw if `x` is a circular structure.

src/index.spec.ts
```diff
@@ -179,6 +186,217 @@ describe('contracts', () => {
...
+  });
+
+  it('dictionary', () => {
+    try {
+      const Check = Dictionary(String);
+      Check.check(null);
+      fail("dictionary passed eventough should've failed");
+    } catch (e) {
+      expect(e.message).toBe('Expected a dictionary [string: string], but was null');
```
pelotom: I would rather just serialize the type here.

>Expected `{ [k: string]: string }` but was `null`.

src/index.spec.ts
```diff
@@ -179,6 +186,217 @@ describe('contracts', () => {
...
+
+  it('dictionary invalid type', () => {
+    try {
+      const Value = Record({ name: String });
+      const Check = Dictionary(Value);
+      Check.check(undefined);
+      fail("dictionary passed eventough should've failed");
+    } catch (e) {
+      expect(e.message).toBe(
+        'Expected a dictionary [string: { name: string; }], but was undefined',
```
pelotom: `[string: { name: string; }]` doesn't parse as TypeScript. Better to just serialize the type:

>Expected `{ name: string }` but was `undefined`.

src/types/union.ts
henrinormak: Any suggestions for an alternative? This is especially useful for cases with API requests etc where it helps to see what the actual value was and not only the expected value.

henrinormak:
Addressed some of the comments, will try to increase the test coverage. For performance discussion, does this help with the discussion - https://github.com/6fold/runtypes/pull/1#issuecomment-381575742

In short, based on a quick performance test the performance seemed to only degrade in the error path (i.e when the check fails) and even then, the slowdown was not that significant (at least in my mind). I suppose, `.check` could be made to work like `.validate` does where it always returns a result and then some wrapper around it could be used for building the error message and throwing that?

henrinormak:
Addressed a couple further things, biggest of which is how Partial works under the hood, instead of using the Union type, we can check against undefined manually. This produces much better error messages for Partial (as Union completely swallows any errors from nested types). This should also be more performant (granted, probably a microscopic difference).

Similarly, I also removed the dependency from Dictionary to Record, in reality the only check Dictionary needed from Record was the one against null/undefined as it does the dictionary/object/array check + key checks later anyway.

I also fixed the code coverage.

pelotom:
> In short, based on a quick performance test the performance seemed to only degrade in the error path 

Yeah, but when checking unions, or heaven forbid unions-within-unions, "errors" can occur many times within a top-level `check`. But based on your benchmarking  I agree the performance penalty doesn't seem too large.

src/types/never.ts
```diff
@@ -8,8 +8,8 @@ export interface Never extends Runtype<never> {
  * Validates nothing (always fails).
  */
 export const Never = create<Never>(
-  () => {
-    throw validationError('Expected nothing but got something');
+  x => {
+    throw validationError(`Expected nothing, but was ${x}`);
```
pelotom: We really need a consistent way of representing the type of an unknown value as a string, preferably in a way which is consistent with `show(Runtype)`. Future work though, I suppose.

pelotom:
Thanks!

henrinormak:
Thanks for merging, any idea when this will be released?

PR #53, opened 2018-05-12
Add Symbol runtype :sparkles:

### LINKED ISSUE #44: Symbol type
Can you add checker for Symbol type?

yuhr:
Resolves #44.


src/types/symbol.ts
```diff
@@ -0,0 +1,21 @@
+import { Runtype, create, validationError } from '../runtype';
+
+export const __global = Function('return this')();
```
pelotom: I don't really understand the need for this. If the goal is to avoid clashing with the `Symbol` class, just use renaming exports/imports as we do with `Array`.

yuhr: Ah sorry, that was just a temporary workaround while I was developing.

src/reflect.ts
```diff
@@ -11,6 +11,7 @@ export type Reflect =
   | { tag: 'boolean' } & Runtype<boolean>
   | { tag: 'number' } & Runtype<number>
   | { tag: 'string' } & Runtype<string>
+  | { tag: 'symbol' } & Runtype<symbol>
```
yuhr: This isn't okay?

pelotom: Ah, sorry, somehow I missed it!

pelotom:
Thanks!

PR #55, opened 2018-05-13
Add branded type support

### LINKED ISSUE #54: `check` should be able to return branded type
The `check` method of a runtype with constraint returns a value of the type without constraint, so sometimes I need to use branded types like this:

```typescript
const A = String.withConstraint(x => x === 'a');
type A = Static<typeof A> & { __brandA: any };
const x: string = 'a';
const a: A = A.check(x); // Type 'string' is not assignable to type 'A'.
const a: A = A.check(x) as A; // OK but looks verbose
```

Then maybe it's good that `check` or some another (defined anew?) method can return some user defined type, for example:

```typescript
let A = String.withConstraint(x => x === 'a');
type A = Static<typeof A> & { __brandA: any };
A = Brand<A>(A);
const x: string = 'a';
const a: A = A.verify(x); // Should be OK
```

I think this is still ugly but absolutely better than doing `as A` whenever `A.check(x)`. 

How do you think?

yuhr:
Resolves #54.


yuhr:
Usage:

```typescript
const A = String.withBrand('A');
type A = Static<typeof A>;

// Also `Brand(name, runtype)` can be used
const B = Brand('B', String.withConstraint(x => /NYANPASU/.test(x)));
type B = Static<typeof B>;

// Merging is possible
const C = Brand('A', String.withBrand('B'));
type C = Static<typeof C>;

// FAIL: A static type-checking error happens here
const a1: A = 'NYANPASU';

// PASS: These are OK
const a2 = 'NYANPASU~' as A;
const a3: A = a2;

// PASS: The type of this should be infered as `A`
const a4 = a2;

try {
  // PASS: No error thrown here, because `Brand` is for static type-checking at all
  const b1: B = B.check('NYANPASU' as A);
  const b2: B = B.check(a4);
  const b3: B = B.check(a4 as C);
} catch (err) {
  // Hence should not reach here
}

// FAIL: These should emit errors
const c1: C = A.check('NYANPASU!');
const c2: C = B.check('NYANPASU!!!') as A;

// PASS: These are possible, because `C` is equivalent to `A & B`
const c3: C = A.check('NYANPASU...') as C;
const c4: C = B.check('NYANPASU.') as A & B;
const c5: C = A.withBrand('B').check('NYANPASU?');
const c6: C = B.withBrand('A').check('NYANPASU~*');
const c7: C = C.withBrand('A').check('*NYANPASU*');
```

package.json
```diff
@@ -21,23 +22,15 @@
...
     "url": "https://github.com/pelotom/runtypes"
   },
   "jest": {
     "verbose": false,
     "testRegex": ".*/*.spec.ts$",
-    "moduleFileExtensions": [
-      "js",
-      "ts"
-    ],
+    "moduleFileExtensions": ["js", "ts"],
```
pelotom: Can you revert this file? I need to add `package.json` to `.prettierignore`.

src/index.spec.ts
```diff
@@ -451,6 +452,12 @@ describe('reflection', () => {
     expectLiteralField(InstanceOf(Test), 'tag', 'instanceof');
     expectLiteralField(Dictionary(Array(InstanceOf(Test))), 'tag', 'dictionary');
   });
+
+  it('brand', () => {
+    const C = Number.withBrand('someNumber') as Brand<'someNumber', Number>;
```
pelotom: Why cast here?

yuhr: That was just nothing... fixed. Thanks 🌟

package.json
yuhr: May I add that?

pelotom:
Thanks!

package.json
pelotom: I'll take care of it :)

src/index.spec.ts
```diff
@@ -40,10 +41,10 @@ type Person = { name: string; likes: Person[] };
 const Person: Runtype<Person> = Lazy(() => Record({ name: String, likes: Array(Person) }));
 
 class SomeClass {
-  constructor(public n: number) {}
+  constructor(public n: number) { }
```
yuhr: because I'm kind of messing up 🤕

pelotom: Well it revealed that Prettier wasn't running correctly in CI, so that's good to know :)

magnushiie:
In your above merging example:
```ts
const C = Brand('A', String.withBrand('B'));
type C = Static<typeof C>;
```
`C` is inferred to be 
```ts
type C = string & {
    [RuntypeName]: "B";
} & {
    [RuntypeName]: "A";
}
```
which is an uninhabitable type - TypeScript currently doesn't compute that to `never` but you could imagine a future version of TypeScript doing that (as currently e.g. `boolean & string` is converted to `never`).

PR #58, opened 2018-07-02
Added tests for Number and fixed Number.check for NaN scenario

elderapo:
Before `Number.check(NaN)` wasn't throwing an error.

### DIFF
--- a/src/types/number.spec.ts
+++ b/src/types/number.spec.ts
@@ -1,0 +1,24 @@
+import { Number } from '..';
+
+describe('number', () => {
+  describe('match', () => {
+    it('works', () => {
+      expect(Number.check(-1)).toEqual(-1);
+      expect(Number.check(-0)).toEqual(-0);
+      expect(Number.check(0)).toEqual(0);
+      expect(Number.check(+0)).toEqual(+0);
+      expect(Number.check(666)).toEqual(666);
+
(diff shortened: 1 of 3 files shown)


pelotom:
This is perhaps a philosophical point, but at least in the type sense, `NaN` _is_ a `number`, however strange it may sound. You can see this in TypeScript by the fact that there is no type error in this statement:

```ts
const x: number = NaN;
```

Since Runtype's `Number` is meant to be a runtime analogue of TypeScript's static `number`, I think it's appropriate as it is. However there's no reason you couldn't define a refinement

```ts
const NotNaN = Number.withConstraint(n => !isNaN(n));
```

elderapo:
I can't really imagine someone using `const n: number = NaN;` on purpose but I get your point.

So instead meaby there could be some "global" setting like:
```
import { treatNaNAsNumber } from 'runtypes';
treatNaNAsNumber(true);
```

or 

```
const numberOrNaN: number = Number.check(NaN, { allowNaN: true }); // allowNaN could be false by default
```

pelotom:
I'd rather not add special case options for things which can be easily built using the core building blocks of the library. You can make your own type which does this using a constraint:

```ts
const StrictNumber = Number.withConstraint(x => isNaN(x) && 'NaN means Not a Number!');
```

PR #59, opened 2018-07-03
Refactored ValidationError and exported it.

elderapo:
I thought it would be a good idea to export `ValidationError` so it will be possible to do something like:
```
import { ValidationError } from 'runtypes';

router.get("/", (req, res) => {
	const someRecord = SomeRecord.check(req.body); // can throw ValidationError

	if (xxx) {
		throw 500 for example...
	}

	...
});

router.on('exception', (ex, req, res) => {
	if (ex instanceof ValidationError) {
		return res.end('validation error');
	}

	return res.end('internal error');
})
```


pelotom:
The validation error wasn't ever really intended to be caught and handled; `check` is for when you really expect the result to be a success and just want the program to blow up if something goes wrong. If you actually want to gracefully handle a validation error the intended approach is to use `validate` instead of `check`, which doesn't throw but returns an object indicating success or failure.

```ts
const result = SomeRecord.validate(req.body);

if (result.success) {
  // ... can use `result.value`, which is of the validated type
} else {
  // ... `result.message` contains the message that would've been in the ValidationError
}
```

elderapo:
Well `SomeRecord.validate(req.body)` doesn't throw an exception so I will have to write boilerplate code every time I try to validate an object.

> `check` is for when you really expect the result to be a success and just want the program to blow up if something goes wrong.

Well it worked that way before this PR and works the same way after this PR. Except that now thrown exception can be properly handled.

soerenBoisen:
In some contexts it leads to much simpler code to defer and centralize error handling by using exceptions instead of if-blocks. I suppose we could always wrap validate and throw an error but why not just expose ValidationError instead and let people choose their error-handling strategy? I don't see the harm, only more flexibility for users of the library.

soerenBoisen:
Another argument for exposing ValidationError: when writing my mocha+chai tests, I sometimes expect validation errors to be thrown. So I might write:
```javascript
expect(() => Recipient.check("loser")).to.throw()
```
This test will pass no matter the error. What I want to write is really
```javascript
expect(() => Recipient.check("loser")).to.throw(ValidationError)
```
to make sure its not some other random error. Of course in this simple example, there isn't much that could go wrong, but I might want to test more complicated code.

pelotom:
Fair enough! If someone wants to fix up the conflicts, I'll merge.

elderapo:
done

PR #71, opened 2018-11-26
Add decorator feature and tests

### LINKED ISSUE #70: Automatic argument checking
Sometimes we need type-check multiples of multiple parameters which are passed from the external world i.e. there're some callbacks to implement, and appending type-checking code for each param to every one of their function bodies is kind of painful. In such case I write a decorator which reduces keypresses more or less and makes functions' logic innate-looking, for example,

```typescript
@checked(Runtype1, Runtype2, Runtype3)
method(param1: Static1, param2: Static2, param3: Static3) {
  ...
}
```

instead of

```typescript
method(param1: Static1, param2: Static2, param3: Static3) {
  Runtype1.check(param1)
  Runtype2.check(param2)
  Runtype3.check(param3)
  ...
}
 ```

How do you think about adding this feature to your library?

yuhr:
Closes #70.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,5 +43,7 @@
     },
     "testEnvironment": "node"
   },
-  "dependencies": {}
+  "dependencies": {
+    "reflect-metadata": "^0.1.12"
+  }
 }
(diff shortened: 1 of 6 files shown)


pelotom:
Thanks, looks good 👍

PR #72, opened 2018-11-27
Fix method decorator function

yuhr:
Sorry I forgot to test return values 😭`@checked` hasn't been calling the original method within the wrapping function when used without `@check`...


src/decorator.ts
```diff
@@ -89,6 +88,7 @@ export function checked(...runtypes: Runtype[]) {
               throw new ValidationError(`${methodId}, argument #${typeIndex}: ${err.message}`);
             }
           });
+          return method.apply(this, args);
```
yuhr: Here

yuhr: Here

PR #74, opened 2018-12-11
Add section to README discussing the variations for handling optional values

kestred:


README.md
```diff
@@ -239,6 +239,49 @@ divide(10, 2); // 5
...
+```ts
+// For variables that might be undefined or null
+const MyString: String;                    // string             (e.g. 'text')
+const MyStringMaybe: String.Or(Undefined); // string | undefined (e.g. 'text', undefined)
+const MyStringNullable: String.Or(Null);   // string | null      (e.g. 'text', null)
+const MyStringVoidable: String.Or(Void);   // string | void      (e.g. 'text', null, undefined)
+```
+
+If a `Record` may or may not have some keys, we can declare the optional
+keys using `myRecord.Partial`.  Partial keys validate successfully if
```
pelotom: I think you meant

```suggestion
keys using `myRecord.And(Partial({ ... }))`.  Partial keys validate successfully if
```

README.md
```diff
@@ -239,6 +239,49 @@ divide(10, 2); // 5
...
+they are absent or undefined (but not null) or the type specified
+(which can be null).
+
+```ts
+// Using `Ship` from above
+const RegisteredShip = Ship.And({
+  // All registered ships must have this flag
+  isRegistered: Literal(true),
+}).Partial({
```
pelotom:
```suggestion
}).And(Partial({
```

README.md
```diff
@@ -239,6 +239,49 @@ divide(10, 2); // 5
...
+  // We may or may not know the ship's classification
+  name: Union(Literal('military'), Literal('civilian')),
+
+  // We may not know the ship's rank (undefined),
+  // we may also know that a civilian ship doesn't have a rank (null)
+  rank: Rank.Or(Null),
+});
```
pelotom:
```suggestion
}));
```

README.md
```diff
@@ -239,6 +239,49 @@ divide(10, 2); // 5
...
+const RegisteredShip = Ship.And({
+  // All registered ships must have this flag
+  isRegistered: Literal(true),
+}).Partial({
+  // We may or may not know the ship's classification
+  name: Union(Literal('military'), Literal('civilian')),
+
+  // We may not know the ship's rank (undefined),
+  // we may also know that a civilian ship doesn't have a rank (null)
+  rank: Rank.Or(Null),
```
pelotom: The comment says "We may not know the ship's rank (undefined) ..." but this code will not allow for `undefined`.

kestred:
Oops, yup!  A few mistakes here and there.  I should have time later in the week for another pass.

README.md
pelotom: I see, yes that makes sense.

pelotom:
Thanks!

PR #80, opened 2019-03-21
Add `isReadonly` flag to Array and Record runtypes

googol:
The flag causes the static type to change. For Record the static type has all properties marked readonly, and for Array the result type is the TS ReadonlyArray (or `readonly T[]` in the upcoming 3.4 version).

Both runtypes also have a convenience method, `asReadonly`, which returns the same type but with the `isReadonly` flag set to true. Some users may prefer the method call over setting the flag for esthetic reasons.


googol:
Hey! I've loved using runtypes for a while now and here's a feature I've wanted. I like to write functional style code with minimal mutations and marking data readonly helps everyone stay on board with that style. It hasn't been possible with runtypes though, so here's a way to enable it.

I hope you agree that this is a useful feature and that I've figured out the tests and all well enough

pelotom:
It seems like the only difference between `Record` and `ReadonlyRecord` (and similarly for `Array`/`ReadonlyArray`) is the static type that they generate, is that correct? In that case I think I'd rather enable this through a more generic facility to alter the static type of a runtype, such as is being discussed in #78, rather than introduce new core runtypes. What do you think?

googol:
Yes you are correct, only the static type is affected

I could see this working as an `.asReadonly` modifier method. However, would it be confusing that it wouldn't work for any other type than `Record` and `Array`.

(Similarly we could have `.asPartial` to replace the `Partial` runtype, but that too would only be applicable to the `Record`.)

 Is that sort of api what you mean, or did I get you wrong?

pelotom:
> I could see this working as an `.asReadonly` modifier method. However, would it be confusing that it wouldn't work for any other type than `Record` and `Array`.

You could put `asReadonly` methods only on the `Array` and `Record` runtypes...

pelotom:
Another way to accomplish this without introducing new core runtypes would just be to overload the `Array` and `Record` runtype constructors with an optional `isReadonly` argument, which changes the static type produced...

googol:
Alright, I actually combined both of those ideas into one, as that seemed like a simple and reasonable thing to do, and prototyped a solution.

I only changed the record and array files to the new format, to see if that style would work for you. If it does, I'll see about all the other files

src/reflect.ts
```diff
@@ -13,7 +13,11 @@ export type Reflect =
   | { tag: 'symbol' } & Runtype<symbol>
   | { tag: 'literal'; value: LiteralBase } & Runtype<LiteralBase>
   | { tag: 'array'; element: Reflect } & Runtype<unknown[]>
+  | { tag: 'readonlyArray'; element: Reflect } & Runtype<ReadonlyArray<unknown>>
   | { tag: 'record'; fields: { [_: string]: Reflect } } & Runtype<{ [_ in string]: unknown }>
+  | { tag: 'readonlyRecord'; fields: { [_: string]: Reflect } } & Runtype<
+      { readonly [_ in string]: unknown }
+    >
```
pelotom: The point of what I was saying was not to introduce new core types like this.

googol: Oh. I didn't fix any of the other files to the new style yet, since I didn't want to invest that effort again before hearing of your opinion on the other implementation.

Please check the record.ts and array.ts files, those are the only ones I updated in my second commit

src/types/array.ts
```diff
@@ -1,16 +1,21 @@
 import { Runtype, Static, create } from '../runtype';
 import { ValidationError } from '../errors';
 
-interface Arr<E extends Runtype> extends Runtype<Static<E>[]> {
+type ArrayStaticType<E extends Runtype, RO extends boolean | undefined> = RO extends true ? ReadonlyArray<Static<E>> : Static<E>[]
+
+interface Arr<E extends Runtype, RO extends boolean | undefined> extends Runtype<ArrayStaticType<E, RO>> {
   tag: 'array';
   element: E;
+  isReadonly: RO;
```
pelotom: I wonder if the flag should actually be persisted beyond determining the type... it doesn’t have any runtime effect so maybe not?

src/types/array.ts
```diff
@@ -1,16 +1,21 @@
...
+
+  asReadonly(): Arr<E, true>;
 }
 
 /**
  * Construct an array runtype from a runtype for its elements.
  */
-function Arr<E extends Runtype>(element: E): Arr<E> {
-  return create<Arr<E>>(
+function Arr<E extends Runtype, RO extends boolean | undefined = false>(element: E, isReadonly?: RO): Arr<E, RO> {
```
pelotom: Is the `| undefined` necessary?

googol: Actually now that I tested it, looks like it isn't required. I had the impression it would be necessary because of the isReadonly parameter is optional

src/types/array.ts
googol: It's needed if we want `show` (and other reflect use cases) to be able to distinguish between the versions.

Is that a reasonable use case?

pelotom: Ah, yeah... I suppose so.

src/types/array.ts
googol: Just using classic function overloading seems to be the best solution actually

googol:
All converted now

googol:
Actually just one more thing popped into my head, maybe adding a positional boolean parameter to the Record/Array constructor isn't the best idea, since in the future you might want to make the Partial type an alias for a modifier on the Record or something else like that. Accumulating more and more positional arguments like that gets very inconvenient.

So probably just having the modifier method for now is the best?

pelotom:
Thanks!

googol:
Thank you!

PR #90, opened 2019-06-03
fix: remove TS 3.5.1 compiler errors

### LINKED ISSUE #89: Incompatible with TypeScript 3.5.1
```
node_modules/runtypes/lib/reflect.d.ts:61:33 - error TS2344: Type 'Runtype<never>' does not satisfy the constraint 'Runtype<unknown>'.
  Types of property 'Or' are incompatible.
    Type '<B extends Runtype<unknown>>(B: B) => Union2<Runtype<never>, B>' is not assignable to type '<B extends Runtype<unknown>>(B: B) => Union2<Runtype<unknown>, B>'.
      Type 'Union2<Runtype<never>, any>' is not assignable to type 'Union2<Runtype<unknown>, any>'.
        Types of property 'match' are incompatible.
          Type 'Match2<Runtype<never>, any>' is not assignable to type 'Match2<Runtype<unknown>, any>'.
            Type 'Matcher2<Runtype<never>, any, Z>' is not assignable to type 'Matcher2<Runtype<unknown>, any, Z>'.
              Type 'Runtype<unknown>' is not assignable to type 'Runtype<never>'.

61     constraint: ConstraintCheck<Runtype<never>>;
                                   ~~~~~~~~~~~~~~

node_modules/runtypes/lib/runtype.d.ts:24:41 - error TS2344: Type 'this' does not satisfy the constraint 'Runtype<unknown>'.
  Type 'Runtype<A>' is not assignable to type 'Runtype<unknown>'.
    Types of property 'Or' are incompatible.
      Type '<B extends Runtype<unknown>>(B: B) => Union2<this, B>' is not assignable to type '<B extends Runtype<unknown>>(B: B) => Union2<Runtype<unknown>, B>'.
        Type 'Union2<this, B>' is not assignable to type 'Union2<Runtype<unknown>, B>'.
          Types of property 'match' are incompatible.
            Type 'Match2<this, B>' is not assignable to type 'Match2<Runtype<unknown>, B>'.
              Type 'Matcher2<this, B, Z>' is not assignable to type 'Matcher2<Runtype<unknown>, B, Z>'.
                Types of parameters 'x' and 'x' are incompatible.
                  Type 'unknown' is not assignable to type 'this["_falseWitness"] & B["_falseWitness"]'.
                    Type 'unknown' is not assignable to type 'this["_falseWitness"]'.
                      Type 'unknown' is not assignable to type 'A'.

24     Or<B extends Runtype>(B: B): Union2<this, B>;
                                           ~~~~

node_modules/runtypes/lib/runtype.d.ts:28:46 - error TS2344: Type 'this' does not satisfy the constraint 'Runtype<unknown>'.

28     And<B extends Runtype>(B: B): Intersect2<this, B>;
                                                ~~~~

node_modules/runtypes/lib/runtype.d.ts:35:51 - error TS2344: Type 'this' does not satisfy the constraint 'Runtype<unknown>'.

35     withConstraint<K>(constraint: ConstraintCheck<this>, args?: K): Constraint<this, K>;
                                                     ~~~~

node_modules/runtypes/lib/runtype.d.ts:35:80 - error TS2344: Type 'this' does not satisfy the constraint 'Runtype<unknown>'.

35     withConstraint<K>(constraint: ConstraintCheck<this>, args?: K): Constraint<this, K>;
                                                                                  ~~~~

node_modules/runtypes/lib/runtype.d.ts:39:53 - error TS2344: Type 'this' does not satisfy the constraint 'Runtype<unknown>'.

39     withBrand<B extends string>(brand: B): Brand<B, this>;
                                                       ~~~~
```

Using Runtypes with TypeScript 3.5.1 emits several errors like above.

justingrant:
Fixes #89. This PR addresses compiler errors when runtypes is compiled using TypeScript 3.5.1. There are 100+ compiler errors when building the runtypes library, but only two root causes. 

Two of the errors are legit type bugs in runtypes code, where we were passing `undefined` or `null` as the first param of the `create()` function, when it should have been a function that returned `undefined` or `null`. 

The rest of the errors are caused by https://github.com/microsoft/TypeScript/issues/31731.  I don't think this is a deliberate breaking change on TS's part. I suspect it's a bug.

Luckily there's an easy workaround we can use to fix runtypes without waiting for TS to fix it.  The problem is in union.ts where we currently declare types like this:  
```ts
export interface Union2<A extends Rt, B extends Rt> extends Rt<Static<A | B>> {
```

To work around that TS issue linked above, the trick is to split up the union, so instead of `Static<A | B>` we should use `Static<A> | Static<B>`.

```ts
export interface Union2<A extends Rt, B extends Rt> extends Rt<Static<A> | Static<B>> {
```

This PR implements this trick throughout union.ts.

### DIFF
--- a/src/types/partial.ts
+++ b/src/types/partial.ts
@@ -16,7 +16,7 @@ export function Part<O extends { [_: string]: Runtype }>(fields: O) {
   return create<Part<O>>(
     x => {
       if (x === null || x === undefined) {
-        const a = create<any>(x, { tag: 'partial', fields });
+        const a = create<any>(x => x, { tag: 'partial', fields });
         throw new ValidationError(`Expected ${show(a)}, but was ${x}`);
       }
 
--- a/src/types/record.ts
+++ b/src/types/record.ts
@@ -27,7 +27,7 @@ export function InternalRecord<O extends { [_: string]: Runtype }, RO extends bo
     create(
       x => {
         if (x === null || x === undefined) {
-          const a = create<any>(x, { tag: 'record', fields });
+          const a = create<any>(x => x, { tag: 'record', fields });
           throw new ValidationError(`Expected ${show(a)}, but was ${x}`);
         }
 
--- a/src/types/union.ts
+++ b/src/types/union.ts
@@ -8,13 +8,14 @@ export interface Union1<A extends Rt> extends Rt<Static<A>> {
   match: Match1<A>;
 }
 
-export interface Union2<A extends Rt, B extends Rt> extends Rt<Static<A | B>> {
+export interface Union2<A extends Rt, B extends Rt> extends Rt<Static<A> | Static<B>> {
   tag: 'union';
   alternatives: [A, B];
   match: Match2<A, B>;
 }
 
-export interface Union3<A extends Rt, B extends Rt, C extends Rt> extends Rt<Static<A | B | C>> {
+export interface Union3<A extends Rt, B extends Rt, C extends Rt>
+  extends Rt<Static<A> | Static<B> | Static<C>> {
   tag: 'union';
   alternatives: [A, B, C];
   match: Match3<A, B, C>;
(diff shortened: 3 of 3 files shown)


justingrant:
BTW, looks like the typescript problem that prompted this PR is expected behavior in TS 3.5+, so we shouldn't expect things to improve without a fix on our end.  See https://github.com/microsoft/TypeScript/issues/31731#issuecomment-498465358.

pelotom:
Awesome, thanks for digging into this.

justingrant:
Cool. Anders replied this morning with an explanation of why this problem happened in 3.5. It's worth a read. https://github.com/microsoft/TypeScript/issues/31731#issuecomment-498691873.  TL;DR - recursive, generic types cause trigger TS to use lowest-common-denominator logic that may be unexpected.

justingrant:
Haha, @pelotom check it out, now Anders is improving Runtypes for us! ;-)

https://github.com/microsoft/TypeScript/issues/31731#issuecomment-498808030

pelotom:
Nice! Yeah those huge overload lists are a legacy of the pre-TS3 world, we can and should do better now.

PR #92, opened 2019-06-21
Changed deprecated Void to Unknown in docs

lkostrowski:


README.md
```diff
@@ -303,7 +303,7 @@ as well as representing keys within records that may or moy not be present.
 const MyString = String;                    // string             (e.g. 'text')
 const MyStringMaybe = String.Or(Undefined); // string | undefined (e.g. 'text', undefined)
 const MyStringNullable = String.Or(Null);   // string | null      (e.g. 'text', null)
-const MyStringVoidable = String.Or(Void);   // string | void      (e.g. 'text', null, undefined)
+const MyStringVoidable = String.Or(Unknown);   // string | void      (e.g. 'text', null, undefined)
```
pelotom: `string | unknown` is the same as `string`. I don’t think this is a useful example anymore, let’s just remove it.

```suggestion
```

PR #96, opened 2019-07-09
Add key to validation Failure Result

ismcodes:
It looks like `check` returns the key at which the type checking failed, so it was helpful (for my own uses) to have `validate` do the same. This way, user feedback with `validate` can be more specific and helpful.

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -26,6 +26,11 @@ export type Failure = {
    * A message indicating the reason validation failed.
    */
   message: string;
+
+  /**
+   * A key indicating the location at which validation failed.
+   */
+  key: string;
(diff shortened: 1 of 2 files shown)


ismcodes:
@pelotom thoughts?

src/result.ts
```diff
@@ -26,6 +26,11 @@ export type Failure = {
    * A message indicating the reason validation failed.
    */
   message: string;
+
+  /**
+   * A key indicating the location at which validation failed.
+   */
+  key: string;
```
pelotom: The `key` should be optional:

```suggestion
  key?: string;
```

Issue #95, opened 2019-07-05
Guard performance...


Hypercubed:
I've found that guards are very slow compared to a base implementation.  In fact, I've seen significant improvement by overriding the guards like:

```ts
String.guard = (x: unknown): x is string => {
  return typeof x === 'string';
};
```

I assume this is due to `guard` calling `validate` which is a try/catch wrapping the `check` function.  Have you thought about reversing the order here so that the `check` function is derived from a fast `guard` instead of the other way around?

Effectively I think this will mean changing the `create` function to be something like:

```ts
export function create<A extends Runtype>(guard: (x: {}) => x is Static<A>, A: any): A;
``` 
instead of:

```ts
export function create<A extends Runtype>(check: (x: {}) => Static<A>, A: any): A;
```

Additionally (or alternatively) could we allow defining the methods of the derived object in the A param:

```ts
export const Number = create<Number>(
  x => { ... }
  { 
    tag: 'number', 
    guard:  (x: unknown): x is string => {
      return typeof x === 'string';
    }
  }
);
```

pelotom:
> I assume this is due to `guard` calling `validate` which is a try/catch wrapping the `check` function. Have you thought about reversing the order here so that the `check` function is derived from a fast `guard` instead of the other way around?

Yeah, others have made this observation and I agree it would be better to build on a validation object rather than exceptions. PRs welcome!

pelotom:
I believe this is resolved now thanks to #99.

PR #99, opened 2019-08-09
optionally pass faster guard function

pabra:
Default guard still uses validate/check.

relates to pelotom/runtypes#95


pelotom:
I think we want to take a more radical approach than this: all runtypes should be specified first and foremost with a guard function, and then `check` should be defined generically in terms of that, not the other way around.

pabra:
Can you please provide an example for an simple type like `Boolean`? Or explain more precisely (best with an example) what you think the goal should look like. 
I have another week where I may have some time to spare and I would love to dedicate it to this great project.
Maybe this discussion should be held in the issue thread #95 :)

pelotom:
Ok, so currently `Boolean` is defined like this:

```ts
export const Boolean = create<Boolean>(
  x => {
    if (typeof x !== 'boolean') throw new ValidationError(`Expected boolean, but was ${typeof x}`);
    return x;
  },
  { tag: 'boolean' },
);
```

But we want to stop throwing `ValidationError` as the base mechanism for type checking, and instead make `create` take a function which returns a validation object (`Result<A>`):

```ts
export const Boolean = create<Boolean>(
  value =>
    typeof value === 'boolean'
      ? { success: true, value }
      : { success: false, message: `Expected boolean, but was ${typeof x}` },
  { tag: 'boolean' },
);
```

Notice that `Runtype` already has a `validate` method on it, and it is defined in terms of the `check` function passed to `create`, which is expected to throw a `ValidationError` in the case of a failed check:

```ts
function validate(value: any): Result<A> {
  try {
    check(value);
    return { success: true, value };
  } catch ({ message, key }) {
    return { success: false, message, key };
  }
}
```

Instead we want to do the opposite: pass in a `validate` function and then define `check` in terms of _it_. Make sense?

pelotom:
Sorry for the delay in getting to this, but it looks great, thanks!

moltar:
That is a _**DRAMATIC**_ change in performance!

https://github.com/moltar/typescript-runtime-type-benchmarks

## Before this PR



## After this PR

pelotom:
Wow, that’s quite amazing!

pabra:
Thanks alot for sharing - that really made my day :)

Zloka:
@moltar I was poking my head around the performance of this library and it seems like you've investigated just yet. However, for some reason the image representing the benchmark after this PR doesn't render for me. Would you be able to share it again? :)

moltar:
Hey @Zloka, sorry, I don't have the image saved.

But what you can do easily is just sift thru the [benchmark repo](https://github.com/moltar/typescript-runtime-type-benchmarks) and find the commit after the upgrade that includes this PR, and one before. Every update gets the data and images saved in the repo every time. So the entire history is there.

moltar:
@Zloka https://github.com/moltar/typescript-runtime-type-benchmarks/search?q=runtypes&type=Commits

Zloka:
@moltar Thanks a lot! :)

Issue #100, opened 2019-08-13
Consider extracting @check / @checked into a separate package


davidje13:
I'm considering using this library, but the runtime dependency (reflect-metadata) seems a bit heavy-handed (declaring new globals and adding properties to Object). It also doesn't look like it's very far along any kind of standardisation process.

I'm not interested in the annotation / decoration features `@check` / `@checked`, and as far as I can tell those are the only reason for this dependency.

I'd like to suggest extracting that feature into a separate package, so that we can use the core type checking without any dependencies, and those who want annotations can use the sub-package.

---

Update: looking at [the code](https://github.com/pelotom/runtypes/blob/master/src/decorator.ts), it looks like the main power of `reflect-metadata` isn't even being used; wouldn't it be possible to achieve the same thing using a `WeakMap` keyed on `target` and drop the dependency entirely?

davidje13:
Closing this since the merged PR resolves my concerns; no need to split the code out any more.

PR #101, opened 2019-08-13
Remove reflect-metadata dependency and simplify decorator code

davidje13:
As noted in #100, the use of `reflect-metadata` is a heavy dependency (due to its use of globals and `object` prototype pollution, and the fact that it is not yet on a standards track and likely to change).

The code wasn't using its power anyway, so I've swapped it out for a `WeakMap` performing the same role. I also tidied up some of the duplicate code and fixed a small mistake in the documentation while I was in the file.

All tests continue to pass, and I have run `yarn format`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,5 @@
     },
     "testEnvironment": "node"
   },
-  "dependencies": {
-    "reflect-metadata": "^0.1.13"
-  }
+  "dependencies": {}
 }
--- a/src/decorator.ts
+++ b/src/decorator.ts
@@ -1,28 +1,40 @@
 import { Runtype } from './runtype';
-
-import 'reflect-metadata';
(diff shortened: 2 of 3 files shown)


davidje13:
I noticed a missing check while changing this but left it out so that this could be a pure refactoring. Specifically: it is possible to annotate the same parameter multiple times.

Also I noticed that it would be very easy to drop `@checked` entirely, with a new syntax like so:

```ts
method2(@check(Runtype1) param1: Static1, param2: any, @check(Runtype3) param3: Static3)
```

This could be done by exploiting the fact that `target[propertyKey]` is the prototype method, meaning it can be wrapped immediately in `@check` (for multiple checked arguments, this would mean wrapping it multiple times, but that should be fine). Maintaining backwards compatibility would be tricky, so maybe for later.

PR #112, opened 2019-10-12
docs for branded types

artalar:
I has an error, but don't know why:

error

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,44 @@ divide(10, 2); // 5
 divide(10, 0); // Throws error: division by zero
 ```
 
+## Branded types
+
+Branded types is a way to emphasize the uniqueness of a type:
+
+```ts
+const Login = String.withBrand('Login');
+const Password = String.withBrand('Password').withConstraint(
+  str => str.length >= 8 || 'Too short password',
+);
+
+const fetchSignIn = Contract(
+  Login,
+  Password,
+  Unknown,
+).enforce((login, pass) => {/*...*/});
+
+const login = Login.check('myLogin');
+const password = Password.check('12345678');
+
+// static type ok, runtime ok
+fetchSignIn(login, password);
+
+// static type error, runtime error
+fetchSignIn(password, login);
+
+// static type error, runtime ok
+fetchSignIn('a@b.c', '12345678');
+```
+
+Branded types are like [opaque types](https://flow.org/en/docs/types/opaque-types) and work as expected, except it is impossible to use branded strings as a key [until we have nominal types](https://github.com/microsoft/TypeScript/pull/33038). Currently the workaround is to add a unique symbol to the main type:
+
+```ts
+const StringBranded = String.withBrand('StringBranded');
+type StringBranded = Static<typeof StringBranded>;
+// Then the type `StringBranded` is computed as:
+// string & { [RuntypeName]: "StringBranded" }
+```
+
 ## Optional values
 
 Runtypes can be used to represent a variable that may be null or undefined
--- a/README.md
+++ b/README.md
@@ -294,40 +294,56 @@ divide(10, 0); // Throws error: division by zero
 
 ## Branded types
 
-Branded types is a way to emphasize the uniqueness of a type:
+Branded types is a way to emphasize the uniqueness of a type. This is useful [until we have nominal types](https://github.com/microsoft/TypeScript/pull/33038):
 
 ```ts
-const Login = String.withBrand('Login');
+const Username = String.withBrand('Username');
 const Password = String.withBrand('Password').withConstraint(
   str => str.length >= 8 || 'Too short password',
 );
 
-const fetchSignIn = Contract(
-  Login,
+const signIn = Contract(
+  Username,
   Password,
   Unknown,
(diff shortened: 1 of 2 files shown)


artalar:
@pelotom can you help me please, what I miss? What the error is mean?

src/runtype.ts
```diff
@@ -85,6 +85,10 @@ export interface Runtype<A = unknown> {
    * Adds a brand to the type.
    */
   withBrand<B extends string>(brand: B): Brand<B, this>;
+  withBrand<B extends string, T extends Static<this>>(
+    brand: B,
+    constraint: ConstraintCheck<this>,
+  ): Brand<B, Runtype<T>>;
```
pelotom: Seems like the overload should just be

```ts
  withBrand<B extends string>(
    brand: B,
    constraint: ConstraintCheck<this>,
  ): Brand<B, this>;
```

What's the rationale for this overload though? Is it really that much of a time saver to be able to write

```ts
Number.withBrand('someNumber').withConstraint(x => x > 3)
```

instead of

```ts
Number.withBrand('someNumber', x => x > 3)
```

? And more importantly, what makes brands special that they should have a shortcut for constraints that other runtypes don't? The nice thing about the `withConstraint` combinator is that it's simple and universal; you don't have to wonder if this or that runtype has such a shortcut. Sometimes conceptual simplicity is worth sacrificing a small amount of concision.

artalar: I think a brand type without constraint checks should not exist: we create a unique type with (I presume) special semantic but don't prove it by a code - looks wrong for me. So when we spread out brand and constraint definitions it pushes us to mistake.

pelotom: >I think a brand type without constraint checks should not exist

A frequent use case for brands that I'm aware of is units of measure or currencies... there's no runtime constraint you can write to validate that a plain `number` represents for example meters instead of degrees fahrenheit.

artalar:
@pelotom I update PR leaving only docs and small bundle size improvement.

src/runtype.ts
artalar: Okay, that's a controversial question, let's leave it

README.md
```diff
@@ -292,6 +292,53 @@ divide(10, 2); // 5
...
+fetchSignIn(login, password);
+
+// static type error, runtime error
+fetchSignIn(password, login);
+
+// static type error, runtime ok
+fetchSignIn('a@b.c', '12345678');
+```
+
+Branded types is like [opaque types](https://flow.org/en/docs/types/opaque-types) and work as expected, except it impossible to use branded strings as a key, [before nominal types release](https://github.com/microsoft/TypeScript/pull/33038), because currently, a brand is setting by tricky workaround by adding unique symbol to the main type:
```
pelotom:
```suggestion
Branded types are like [opaque types](https://flow.org/en/docs/types/opaque-types) and work as expected, except it is impossible to use branded strings as a key [until we have nominal types](https://github.com/microsoft/TypeScript/pull/33038). Currently the workaround is to add a unique symbol to the main type:
```

README.md
```diff
@@ -292,6 +292,53 @@ divide(10, 2); // 5
...
+```ts
+const StringBranded = String.withBrand(
+  'my brand',
+)
+type StringBrandedStatic =  Static<typeof StringBranded>
+
+type Prop = { [K in StringBrandedStatic]: 'prop' }[
+  // `Type 'string & { [RuntypeName]: "my brand"; }' cannot be used as an index type.`
+  StringBrandedStatic
+]
```
yuhr, line 339:
```suggestion
const StringBranded = String.withBrand('StringBranded');
type StringBranded = Static<typeof StringBranded>;
// Then the type `StringBranded` is computed as:
// string & { [RuntypeName]: "StringBranded" }
```

yuhr (changes requested):
Sorry for my laziness of adding docs in spite of implementing `Brand`, I really appreciate this PR. Thanks @artalar!

Here're small points:

- Don't forget removing "optional constraint" part
- I don't think `withBrand` needs new lines inside of parens; there's just a single argument
- To be consistent, use semicolons in the second code block

I don't know whether my realization is correct actually about what the second code block is for, but my suggestion may be more clear in order to illustrate the structure of branded types.

Optionally, I'd recommend using **the exact same name** for runtype, static type and branded RuntypeName. This makes search & replace quite easy, so worth documented as kind of best practice, I think.

yuhr:
By the way, I found [ryansmith94/rulr](https://github.com/ryansmith94/rulr) has pretty similar feature with runtypes. I prefer (and would recommend 😜) runtypes' succinct and satisfactory APIs, but as for type branding the lack of documentation can be pretty easily misunderstood as a lack of feature. I hope this gets merged!

artalar:
> Here're small points:

I will do it today, sorry for my disappearing.

artalar:
done ^

yuhr:
Sorry I've noticed that I've misunderstood the "caveats" of branded types in your original PR, but I can't confirm the workaround you provided is working, at least in TS 4.2.2. So let me add some other commits...

yuhr (approved):

PR #113, opened 2019-10-13
implement inline optional for record

artalar:
The better way, then `Record({}).And(Partail({}))`


pelotom:
I like the motivation behind this, but it changes the semantics slightly from TypeScript:

```ts
{ foo?: number }
```

is not the same as

```ts
{ foo: number | undefined }
```

The former says that `foo` can be present or not in the object, and if it is present it could be `undefined`. The latter says that a `foo` property must always be present, but its value could be undefined.

To do this the right way, I think we would need the `Optional` property not to be a Runtype, but instead a special branding type for use only in `Record` values (it doesn't make sense to talk about `Optional(Number)` as a type in isolation). And even then I'm not sure it's possible; it seems like it would require some type crazy syntax that doesn't currently exist e.g.

```ts
{
  [K in keyof O][O[K] extends Optional ? +? : -?]: O[K] extends Optional<infer T> ? T : O[K]
}
```

artalar:
> it doesn't make sense to talk about Optional(Number) as a type in isolation
Totally agree, but I did not find other ways...

> And even then I'm not sure it's possible
But I already do it :)
example

The question is how the right API may look like... Current workaround by Record and Partial interaction looks weird(

pelotom:
I think it's reasonable to make an alias for the `T | undefined` type, understanding that it's not the same as an optional property on a record. It doesn't need to be a new core runtype though; we could simply add a new unary operator on the `Runtype` interface, similar to `Or` and `And`:

```diff
  /**
   * Union this Runtype with another.
   */
  Or<B extends Runtype>(B: B): Union2<this, B>;

  /**
   * Intersect this Runtype with another.
   */
  And<B extends Runtype>(B: B): Intersect2<this, B>;

+ /**
+  * 
+  */
+ Optional: Runtype<Static<this> | undefined>;
```

implemented as

```diff
A.Or = Or;
A.And = And;
+A.Optional = Union(A, Literal(undefined));
```

yuhr:
> it doesn't make sense to talk about Optional(Number) as a type in isolation

Some languages have an algebraic datatype called `Option`, so I don't think it's nonsense to use this wording for an alias of unionizing with `undefined`, even in isolation.

I'd go for the OP's approach, which is having `Optional` as a discrete runtype. My proposal is:

1. If `Optional` runtype is used in isolation, it works just as an alias for `Number.Or(Undefined)` whose static type is `number | undefined`
2. If used in a `Record`, it works as genuinely an optional property, with its static type resulting in `{ foo?: number }`
3. If one wants to really get `{ foo: number | undefined }`, an explicit unionization i.e. `Number.Or(Undefined)` is needed

Why we need it to be a discrete runtype is nothing else but the point 2. `Record` runtype should *dynamically* check if a field is `Optional` or not, so that it can be properly handled.

yuhr:
Anyway, the PR branch is a bit stale to merge, so I'm going to update it.

yuhr:
I ended up having the shorthand syntax of `Optional` as like `String.optional()`. This is not much shorter than writing `Optional(String)` when you import `Optional` directly from `runtypes`, but if you use scoped import i.e. `import * as rt from "runtypes"`, it would look better to write `rt.String.optional()` rather than `rt.Optional(rt.String)`. I decided to decapitalize as per consistency with upcoming APIs such as `.pick()` and `.omit()` in #161 (I think also `.And()` and `.Or()` should be renamed and deprecated, off-topic though).

artalar:
Awesome 👍

mushegha:
Nice addition, definitely useful. When it's hitting npm?

yuhr:
Sorry for being late, it should be available in `v5.2.0` now!

README.md
```diff
@@ -346,42 +346,40 @@ type SomeMap = Map<StringBranded, number>;
...
     // we may also know that a civilian ship doesn't have a rank (e.g. null)
-    rank: Rank.Or(Null),
+    rank: Optional(Rank.Or(Null)),
   }),
 );
 ```
 
-If a record has keys which _must be present_ but can be null, then use
-the `Record` runtype normally instead.
+There's a difference between `Union(String, Undefined)` and `Optional(String)` iff they are used within a `Record`; the former means "_**it must be present**, and must be `string` or `undefined`_", while the latter means "_**it can be present or missing**, but must be `string` or `undefined` if present_".
```
OmgImAlexis, line 380: typo on `iff`

Issue #114, opened 2019-10-13
Contract for async function


artalar:
How can I describe the result of an async function?

For example:

```ts
const fetchSignIn = t
  .Contract(
    Login,
    Password,
    // best place for type declaration
    // here we need possibility to describe runtype `SignInResponse`
    t.InstanceOf(Promise)
  )
  .enforce(async (login, password) => {
    const response = await request('/sign-in', {login, password})
    // inline `check` unreliably if we has a few result branches 
    return SignInResponse.check(response)
  });
```

pelotom:
I'm not sure how this might work. Open to ideas...

artalar:
May be

![example](https://user-images.githubusercontent.com/27290320/67399721-54acb080-f5b5-11e9-9b26-e91488f1e9d0.png)

pelotom:
Hm, it's not a runtype though, it's a type of contract...

artalar:
I mean when we call `check` with a promise - it adds `then(fulfilled.check)`.
With these solutions we may describe contract so:
```ts
const fetchSignIn = t
  .Contract(
    Login,
    Password,
    t.Async(SignInResponse)
  )

```

zhaoyao91:
Oops, AsyncContract is great. I just about leaving when suddenly found this issue. Why not add it to the readme document?

PR #122, opened 2020-01-31
Created asyncContract

schicks:
Implementation of the AsyncContract sketched out as a potential solution for #114.
I didn't see any contributing guidelines; is this reasonable in terms of style/content as a PR?

### DIFF
--- a/src/asynccontract.spec.ts
+++ b/src/asynccontract.spec.ts
@@ -1,0 +1,16 @@
+import { AsyncContract, Number } from '.';
+
+describe('AsyncContract', () => {
+  describe('when function does not return a promise', () => {
+    it('throws a validation error', () => {
+      const contractedFunction = AsyncContract(Number).enforce(() => 7 as any);
+      expect(contractedFunction).toThrow();
+    });
+  });
+  describe('when a function does return a promise', () => {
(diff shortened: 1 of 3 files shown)


src/asynccontract.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  describe('when function does not return a promise', () => {
+    it('throws a validation error', () => {
+      const contractedFunction = AsyncContract(Number).enforce(() => 7 as any);
+      expect(contractedFunction).toThrow(ValidationError);
+    });
+  });
+  describe('when a function does return a promise, but for the wrong type', () => {
+    it('throws a validation error asynchronously', async () => {
+      const contractedFunction = AsyncContract(Number).enforce(() => Promise.resolve('hi' as any));
+      await expect(contractedFunction()).rejects;
```
pelotom: I don't think this line tests what you want it to. I'm not sure if there's an easier way to do it with Jest's matchers, but this would get the job done:

```suggestion
      try {
        await contractedFunction();
        fail();
      } catch (e) {
        expect(e).toBeInstanceOf(ValidationError);
      }
```

schicks: That definitely seems clearer regardless.

Issue #115, opened 2019-11-07
Recursive Runtypes can lead to stack overflows


mscharley:
I have the following Runtype which I use as part of an ElasticSearch query:

```typescript
import { Dictionary, Lazy, Literal, Number, Partial, Record as RtRecord, Runtype, String, Union } from "runtypes";

export interface ElasticSearchFieldMapping {
  type: "boolean" | "date" | "double" | "integer" | "keyword" | "text";
  analyzer?: string;
  fields?: Record<string, ElasticSearchFieldMapping>;
  ignore_above?: number;
}

export const ElasticSearchFieldMapping: Runtype<ElasticSearchFieldMapping> = Lazy(() =>
  RtRecord({
    type: Union(
      Literal("boolean"),
      Literal("date"),
      Literal("double"),
      Literal("integer"),
      Literal("keyword"),
      Literal("text"),
    ),
  }).And(Partial({
    analyzer: String,
    fields: Dictionary(ElasticSearchFieldMapping, "string"),
    ignore_above: Number,
  })),
);
```

When I use `.check(input)` to throw an exception, the code that tries to describe the expected input ends up in a stack overflow:

```
RangeError: Maximum call stack size exceeded
    at Array.map (<anonymous>)
    at /opt/nodejs/node_modules/runtypes/lib/show.js:29:22
    at Array.map (<anonymous>)
    at /opt/nodejs/node_modules/runtypes/lib/show.js:44:56
    at /opt/nodejs/node_modules/runtypes/lib/show.js:24:61
    at /opt/nodejs/node_modules/runtypes/lib/show.js:36:80
    at Array.map (<anonymous>)
    at /opt/nodejs/node_modules/runtypes/lib/show.js:36:31
    at Array.map (<anonymous>)
    at /opt/nodejs/node_modules/runtypes/lib/show.js:44:56
```

pelotom:
What is `input` in your example? Everything I've tried seems to work fine.

mscharley:
Anything that fails in general but at least anything that fails on the recursion boundary should trigger it. It must be a failure, checking works fine it’s only generating the string message for the Error that doesn’t work in my experience. I’m out today but if you can’t get a repro in the next 24h I’ll put something together when I’m at work next.

warpr:
Hello, I'm having this problem as well.  For me checking in itself seems to hit that RangeError:

```
> $ node --version
v12.14.0
```

```
> $ node --experimental-modules ./demo.mjs
[...]

PR #123, opened 2020-02-01
added cycle breaking sets to InternalRecord

schicks:
Potential way to address #115. However, I don't know this code very well and would appreciate a close eye to check that what I'm doing here doesn't have unforeseen consequences uncovered by the tests.

### DIFF
--- a/src/runtype.ts
+++ b/src/runtype.ts
@@ -167,25 +167,21 @@ type VisitedState = {
 function VisitedState(): VisitedState {
   const members: WeakMap<object, WeakMap<Runtype, true>> = new WeakMap();
 
-  const returned = {
-    add: (candidate: object, type: Runtype) => {
-      if (candidate === null || !(typeof candidate === 'object')) return returned;
-      const typeSet = members.get(candidate);
-      members.set(
-        candidate,
-        typeSet
-          ? typeSet.set(type, true)
-          : (new WeakMap() as WeakMap<Runtype, true>).set(type, true),
-      );
-
-      return returned;
-    },
-    has: (candidate: object, type: Runtype) => {
-      const typeSet = members.get(candidate);
-      const value = (typeSet && typeSet.get(type)) || false;
-      returned.add(candidate, type);
-      return value;
-    },
+  const add = (candidate: object, type: Runtype) => {
+    if (candidate === null || !(typeof candidate === 'object')) return;
+    const typeSet = members.get(candidate);
+    members.set(
+      candidate,
+      typeSet ? typeSet.set(type, true) : (new WeakMap() as WeakMap<Runtype, true>).set(type, true),
+    );
   };
-  return returned;
+
+  const has = (candidate: object, type: Runtype) => {
+    const typeSet = members.get(candidate);
+    const value = (typeSet && typeSet.get(type)) || false;
+    add(candidate, type);
+    return value;
+  };
+
+  return { has };
 }
--- a/src/runtype.ts
+++ b/src/runtype.ts
@@ -31,7 +31,7 @@ export interface Runtype<A = unknown> {
   /**
    * Validation function used internally that additionally takes sets to track cycle breaking
    */
-  innerValidate(x: any, visitedSet: VisitedState, failedSet: VisitedState): Result<A>;
+  innerValidate(x: any, visited: VisitedState): Result<A>;
 
   /**
    * A type guard for this runtype.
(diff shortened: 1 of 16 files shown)


schicks:
That's a good point. I'm having a little trouble finding a place in the code that would stay in context for the entirety of an invocation of validate (including all recursive calls). Ideally, we would wrap something on the `Lazy` type with this logic, but similarly it wasn't clear to me where that would go. Could you point me to the right place in the code, or sketch out a test that would prove that there is not mutable state attached to the type itself?

schicks:
This version involved modifying the signature of validate a little which I was nervous about, but all existing validate functions still meet the new type. Now the sets are created on any validate call where they are not passed as arguments, and validate functions which themselves call some validate function should pass in the sets to the inner calls. I only implemented it on Record and Array for now because that's enough to get the tests to pass,but if this seems like a good approach I'll go back through and add it to any other recursively validating types.

schicks:
To have test coverage on the set usage in the array validator we need an additional test;

```typescript
type GraphNode = GraphNode[]; // graph nodes are just arrays of their neighbors
const GraphNode: Runtype<GraphNode> = Lazy(() => Array(GraphNode));
type Graph = GraphNode[];
const Graph: Runtype<Graph> = Array(GraphNode);
const nodeA: GraphNode = [];
const nodeB: GraphNode = [nodeA];
nodeA.push(nodeB);
const barbell: Graph = [nodeA, nodeB];
```

However, the `GraphNode` type isn't valid typescript until 3.7. Is it reasonable to bump the dev version of typescript for this?

src/types/array.ts
```diff
@@ -21,17 +21,20 @@ function InternalArr<E extends Runtype, RO extends boolean>(
 ): Arr<E, RO> {
   return withExtraModifierFuncs(
     create(
-      xs => {
+      (xs, visitedSet = new Set(), failedSet = new Set()) => {
```
pelotom: Shouldn't need to create defaults here, all arguments should be supplied for you.

schicks: I'm not sure I understand. How can we know that the defaults will be supplied? for instance, in the graph test case wouldn't this validate function be called at the top level?

pelotom: This is not the `validate` function, this is the argument to `create`, which is an internal utility. We can guarantee that the function passed to `create` always has its arguments supplied.

schicks: Ok, so we would modify the create function in `runtypes.ts` to wrap the passed in function in something that provides it with empty sets and assign the wrapping function to `validate`?

schicks: I think 77eed40 should address this, if I've understood you correctly.

schicks:
Just to confirm, the remaining types to augment are;

- [ ] tuple
- [ ] partial

Is that right?

schicks:
I think this covers every type that is not a base case now. I am a little worried about partial; I don't think that test is thorough enough, it seems like we should need to add to the sets in a few cases there but I can't figure out how to write a test that will reveal that problem. Does that sound right to you?

schicks:
There is a finicky problem that I'm stuck on related to intersection. If we take the intersection of a variety of normal, straightforward records, it is important to remove the object being checked from the visited set between checks of each part of the intersection, so that we don't erroneously claim that something is successful because it succeeded on some other part of the intersection. however, once we do this types like `Ambi` have the infinite recursion problem again;
```typescript
type Ambi = { left: Ambi } & { right: Ambi };
const Ambi: Runtype<Ambi> = Lazy(() => Intersect(Record({ left: Ambi }), Record({ right: Ambi })));
const ambi: Ambi = { left: (null as any) as Ambi, right: (null as any) as Ambi };
ambi.left = ambi;
ambi.right = ambi;
```
I'm not sure how to solve that.

schicks:
I guess we could build a custom state that held tuples of `(Runtype, candidate) ` so that we maintain a record specifically of whether a candidate has visited a type. I'll try that tomorrow.

schicks:
I think this should be a minimum working state.

src/runtype.ts
```diff
@@ -99,9 +104,13 @@ export interface Runtype<A = unknown> {
...
 
-export function create<A extends Runtype>(validate: (x: any) => Result<Static<A>>, A: any): A {
+export function create<A extends Runtype>(
+  validate: (x: any, visited: VisitedState, self: A) => Result<Static<A>>,
+  A: any,
+): A {
   A.check = check;
-  A.validate = validate;
+  A.innerValidate = (x: any, visited: VisitedState) => validate(x, visited, A);
+  A.validate = (x: any) => validate(x, VisitedState(), A);
```
pelotom: It doesn't seem like we need to make every single runtype responsible for making these checks, we can just do it centrally here:

```ts
  A.validate = (x: any, visited: VisitedState = VisitedState()) => {
    if (visited.has(x, A)) return { success: true, value: x };
    return validate(x, visited);
  };
```

I also don't think `innerValidate` is useful, we just need an optional extra parameter on `validate`.

schicks: That's true. I was thinking that we couldn't do that because of Union, but it looks like while Union doesn't need the cycle breaking check it doesn't suffer from having it.

schicks:
I'm a little unsure about having the visitedState as that optional parameter. It seems like that isn't something that external users of the runtype should be aware of or poking at. The advantage of something like `innerValidate` was that it separated it off from the current API, but obviously it would be better to mark something like that explicitly as private. 

Basically my concern is that looking ahead to addressing #8, it seems like that will also need to change the internal behavior of validate and build up some state in a similar way. Following this pattern seems like it might make `validate`'s signature very ugly after a few more of these.

pelotom:
> I'm a little unsure about having the visitedState as that optional parameter. It seems like that isn't something that external users of the runtype should be aware of or poking at. The advantage of something like `innerValidate` was that it separated it off from the current API, but obviously it would be better to mark something like that explicitly as private.

Yeah, there's not really a way to make it truly private. I would suggest giving it a `_`-prefixed name, leaving it off the public `Runtype` interface, and have a helper function exported from the module, e.g.

```ts
export function innerValidate(A: Runtype, value: any, visited: VisitedState) {
  return (A._innerValidate as any)(value, visited);
}
```

then use this helper in the individual runtype definitions.

schicks:
Is this in a good state now @pelotom ? I'd like to work off of this as a baseline for a few other issues but was waiting for it to be merge ready.

Issue #118, opened 2020-01-05
Document .asReadonly()


lirbank:
https://github.com/pelotom/runtypes/pull/80 implements `.asReadonly()` which is awesome! Would be great to have it mentioned in the README so the next dev doesn't miss it.

cc @googol

Cheers!

pelotom:
A PR to document this would be welcome!

PR #126, opened 2020-02-02
Add documentation for readonly records and arrays

googol:
Fixes https://github.com/pelotom/runtypes/issues/118


pelotom:
Thanks!

PR #131, opened 2020-03-10
Add Struct and Maybe types

yosbelms:
`Struct` is an extension of `Record` that sets optional properties of `Maybe` type.

```ts
const User = Struct({
  id: Maybe(Number),
  name: String,
})
```

Equivalences:

`Maybe(T)` == `Union(T, Undefined)`
`Struct({ opt: Maybe(O), req: R })` == `Record({ req: R }).And(Partial({ opt: O }))`

Additionally the `eval` utility does the same that `check` but also check types at compile time.
```ts
const user = User.eval({
  id: 'invalid'
})

// Error...
```

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -29,6 +29,8 @@ import {
   InstanceOf,
   Brand,
   Guard,
+  Maybe,
+  Struct,
 } from './index';
 
 import { Constructor } from './types/instanceof';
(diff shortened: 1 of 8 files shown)


ben-pr-p:
Definitely like the `key: Maybe(O)` over `.and(Partial(...`!

yosbelms:
@ben-pr-p if you want to try it the npm package is `yosbelms-runtypes`

nebur395:
Hey guys,
Awesome repository by the way, thank you for your work.
I'd really love something like this. Is it possible to boost the review of this PR?
Thank you :)

pelotom:
This is an interesting approach. Why is the `Struct` type needed, instead of just augmenting `Record`?

yuhr:
Ping @yosbelms. I'm wondering the same thing too. I'm very keen to merge `Maybe` only.

yuhr:
Let's close in favor of #113, where the problem is better explained.

PR #134, opened 2020-03-14
Fix ValidationError message bug

richgieg:
Previously the `ValidationError` constructor was using the `public` modifier on the `message` parameter which redefined the existing `message` field inherited from `Error`. The `message` field was being re-initialized back to the value of the `message` argument after the `super` call.

Original JavaScript Output:

```
export class ValidationError extends Error {
    constructor(message, key) {
        super(key ? `${message} in ${key}` : message);
        this.message = message;
        this.key = key;
        this.name = 'ValidationError';
        Object.setPrototypeOf(this, ValidationError.prototype);
    }
}
```

Removing the `public` modifier fixes this issue. The updated JavaScript output no longer has the `this.message = message` line after the `super` call.

Updated JavaScript Output:

```
export class ValidationError extends Error {
    constructor(message, key) {
        super(key ? `${message} in ${key}` : message);
        this.key = key;
        this.name = 'ValidationError';
        Object.setPrototypeOf(this, ValidationError.prototype);
    }
}
```

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -1,7 +1,7 @@
 export class ValidationError extends Error {
   public name: string = 'ValidationError';
 
-  constructor(public message: string, public key?: string) {
+  constructor(message: string, public key?: string) {
     super(key ? `${message} in ${key}` : message);
 
     Object.setPrototypeOf(this, ValidationError.prototype);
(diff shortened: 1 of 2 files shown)


seemk:
Having the same issue, hope this gets merged.

milanij:
Yes, here's another vote for merging this in!

Would be really helpful to see where the errors are happening.

milanij:
@pelotom Is there any harm to merging this in? Seems like a bug that should get fixed. Thanks! 😄

pelotom:
Thanks, looks good!

PR #138, opened 2020-03-22
Fix for error TS2349

jakajancar:
```
error TS2349: This expression is not callable.
  Type 'never' has no call signatures.

► http://localhost:8000/raw.githubusercontent.com/pelotom/runtypes/master/src/show.ts:65:9

65   throw Error('impossible');
           ~~~~~
```

### DIFF
--- a/src/show.ts
+++ b/src/show.ts
@@ -58,11 +58,12 @@ const show = (needsParens: boolean, circular: Set<Reflect>) => (refl: Reflect): 
         return `InstanceOf<${name}>`;
       case 'brand':
         return show(needsParens, circular)(refl.entity);
+      default:
+        throw Error('impossible');
     }
   } finally {
     circular.delete(refl);
   }
-  throw Error('impossible');
 };
 
 export default show(false, new Set<Reflect>());


pelotom:
I'm not sure how you're trying to build the library but if you run `yarn && yarn build` it should work.

jakajancar:
I’m not building it, I use it with Deno which supports TS natively.

pelotom:
> I’m not building it, I use it with Deno which supports TS natively.

Apparently not fully, if it's reporting errors when `yarn build` does not. Does Deno not support exhaustiveness checking?

jakajancar:
It’s using the official tsc. Maybe the default tsconfig is different?

pelotom:
> It’s using the official tsc. Maybe the default tsconfig is different?

You would need to make it use the tsconfig in the root of this repo.

jakajancar:
Yeah, but changing one's project tsconfig (or, in my case, adding it) because of one dependency... :/ Would be certainly nicer if the code was just fixed to work with stricter settings.

Either way, understood if Deno is not your target, you don't really care. Without the proper tests, also who knows what other problems could accidentally be introduced down the line.

👍

pelotom:
I don't know Deno, but if it doesn't allow for dependencies to be compiled under different configuration than your own project that seems like a pretty severe design flaw IMO (one which regular TypeScript does not suffer from). BTW, this project _is_ using the stricter setting: by not using a default case and instead relying on exhaustiveness checking to prove that the function is still total, it protects from the situation where you add another case but fail to handle it in the switch statement. If you add another case to `Reflect`, you will suddenly have a type error in that function because it is not exhaustive.

PR #140, opened 2020-03-25
Fix some code examples in the readme

flawyte:


pelotom:
What's wrong with the word "coordinates"?

flawyte:
Nothing wrong with the word itself, it's just that all other code snippets through the README use `location` and have no `coordinates`, so I thought it was a mistake.

pelotom:
Ah, good point!

PR #141, opened 2020-04-15
Fix for error TS2349: This expression is not callable.

jakajancar:
Fix for error TS2349: This expression is not callable. Type never has no call signatures.


jakajancar:
Eh, noticed I already submitted this.

PR #144, opened 2020-04-28
Provide a single-file browser-friendly distribution

madnl:
The emitted code of the library contains circular references within modules. Some browser bundler tools do not support this which can hinder adoption.

This patch adds an alternative way of distributing the library tailored for this use case. Using webpack, a new way of distributing the library in the form of a single-file module without any external dependencies is generated in the `dist/runtypes.js`.

The new output does not replace the current emitted code, but instead it is generated in addition to it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Compiled output
 lib/
+dist/
 
 # Code coverage output
 coverage/
--- a/package.json
+++ b/package.json
@@ -5,7 +5,9 @@
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
   "scripts": {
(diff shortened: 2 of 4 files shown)


pelotom:
I would rather not add a dependency on a particular bundler to this library. Is there a way to re-do the imports so as not to confuse your bundler?

yuhr:
@madnl What are "Some browser bundler tools" actually? Could you provide some repros for them?

madnl:
> @madnl What are "Some browser bundler tools" actually? Could you provide some repros for them?

Unfortunately no. It's an internal building tool. But I think it's not reasonable to expect refactoring the dependencies inside this package for that. I'm going to close the issue.

Issue #67, opened 2018-10-17
Add a TaggedUnion runtype.


mscharley:
There's a fairly common pattern for using union types called tagged unions. Using the current `Union` runtype for these is functionally complete, however when a `.check()` fails, it's usually near impossible to work out why. I'd like to propose a new `TaggedUnion` runtype. You could use it as follows:

```typescript
const AB = TaggedUnion("tag", {
  a: Record({ aprop: String }),
  b: Record({ bprop: Number }),
});

AB.check({ tag: "a", aprop: "foo" }); // success.
AB.check({ tag: "b", bprop: 10 }); // success.
AB.check({ tag: "c", cprop: "bar" }); // Error: Unknown tag: Expected "a" | "b", but was "c"
AB.check({ tag: "a", bprop: 10 }); // Error: Expected { tag: "a"; aprop: string; }, but was object.

AB.checkTag({ tag: "a" }); // true
AB.check({ tag: "a" }); // Error: Expected { tag: "a"; aprop: string; }, but was object.
AB.checkTag({ tag: "c", cprop: "bar" }); // false
```

Having a seperate `checkTag` is often useful for deserialising input from changing upstream providers such as webhooks where they may add new event types which you want to discard or log differently because you typically care less about them, but you do care if you get an event you thought you knew about which has a structure other than you were expecting.

As an example of why I feel this is necessary, the current approach leaves me with errors like the following while trying to deserialise Mailgun webhooks (if there is an error):

[...]

PR #149, opened 2020-06-05
Implement Discriminated Unions

ivank:
First of all, AWESOME library! One thing has been bugging me about it though and its support for Discriminated Unions.

This is my attempt on addressing it, and I've seen similar issues being raised too.

So with this PR, when those conditions are met:
- all the alternatives of a union are records
- they have at least one field in common, that is a literal in each
record with unique values

Then when checking the alternatives we can pick the correct one first, and then perform the validation, thus producing more specific error messages.

This is something already present in TypeScript itself https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions and seems to have been requested before https://github.com/pelotom/runtypes/issues/67 .
Though this implementation would not work in all the cases covered by TS, but it's a start.

I'll be glad to add more tests or work on the implementation on your direction. Sorry for the long implementation, but the you seem to be supporting way older version of JS than I'm used to so had to implement the logic in for loops and such.

### DIFF
--- a/src/types/union.ts
+++ b/src/types/union.ts
@@ -1,5 +1,7 @@
 import { Runtype as Rt, Static, create, innerValidate } from '../runtype';
 import show from '../show';
+import { LiteralBase } from './literal';
+import { hasKey } from '../util';
 
 export interface Union1<A extends Rt> extends Rt<Static<A>> {
   tag: 'union';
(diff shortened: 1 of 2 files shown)


src/types/union.ts
```diff
@@ -966,6 +968,46 @@ export function Union(...alternatives: Rt[]): any {
 
   return create(
     (value, visited) => {
+      if (
+        alternatives.length > 1 &&
```
pelotom, line 972: Why is this condition necessary?

ivank: I wouldn’t call it necessary, just a performance optimization that we can easily get rid off. If you think its not worth it I can remove it for sure. It might help with coverage as well. 

Also we can implement some of the checks with type guards and that would help with coverage as well, but I figured I’d use existing types as much as possible. The current way Reflect is implemented I can’t see a way to reuse the types for typeguards.

pelotom: Yeah, I think for simplicity's sake we should remove it.

kael-shipman:
Hey all! Very interested in seeing this come through :). This is a godsend library, and I love to see it improving like this.

PR #150, opened 2020-06-06
Update Type guards URL reference

PabloSzx:
The previous URL in the README doesn't work anymore and it just redirects to the home page


pelotom:
Thanks!

PR #153, opened 2020-07-14
Merge Record and Partial

### LINKED ISSUE #152: Any reason for missing asReadonly on Partial?
`asReadonly()` only seems to exist on `Array` and `Record` types. Would it be easy to extend this to support `Partial` as well? e.g. I'd like to be able to do:
```typescript
Record({
  required: Number
}).asReadonly().And(Partial({
  optional: Number
}).asReadonly());

// -> {readonly required: number;} & {readonly optional?: number | undefined;}
```

keithlayne:
This seems to work, but I have some questions/concerns:

- The test stuff is completely arbitrary and I don't know if what I've done really follows the intent
- Not sure I named things the way you like - I decided to just be symmetric with `Record`
- Now `partial.ts` is basically a carbon copy of `record.ts`...it's just begging to be refactored 

Thoughts?

Fixes #152


pelotom:
> * The test stuff is completely arbitrary and I don't know if what I've done really follows the intent

What do you mean by arbitrary?

> * Now `partial.ts` is basically a carbon copy of `record.ts`...it's just begging to be refactored

I agree, what do you think about making `Record` take an optional argument which marks it as partial, and then `Partial` just being an alias to `Record` with that flag set?

keithlayne:
Sorry for being unclear.  I meant what *I* wrote in the tests was pretty much just guessing.  Obviously `readonly` has no impact on runtime, so (I guess) it should have the same tests as `Partial`.  Soooo...I didn't know if you'd want me to do something better in the tests.

I will give the refactor a go, looks pretty simple.  Thanks for responding so quickly!

src/types/record.ts
```diff
@@ -2,26 +2,52 @@ import { Runtype, Static, create, innerValidate } from '../runtype';
...
-type RecordStaticType<O extends { [_: string]: Runtype }, RO extends boolean> = RO extends true
+type RecordStaticType<
+  O extends { [_: string]: Runtype },
+  Part extends boolean,
+  RO extends boolean
+> = Part extends true
+  ? RO extends true
+    ? { readonly [K in keyof O]?: Static<O[K]> }
+    : { [K in keyof O]?: Static<O[K]> }
+  : RO extends true
```
keithlayne, line 13: This is gross, but there's no better way I can think of using booleans.

src/types/record.ts
```diff
@@ -2,26 +2,52 @@ import { Runtype, Static, create, innerValidate } from '../runtype';
...
   ? { readonly [K in keyof O]: Static<O[K]> }
   : { [K in keyof O]: Static<O[K]> };
 
-export interface Record<O extends { [_: string]: Runtype }, RO extends boolean>
-  extends Runtype<RecordStaticType<O, RO>> {
+interface InternalRecord<
```
keithlayne, line 17: Decided to rename this and maintain the old `Record`` interface.  What do you think?

Also not exporting, I *think* TS would give an error about private dependencies if that were an issue.

src/types/record.ts
```diff
@@ -2,26 +2,52 @@ import { Runtype, Static, create, innerValidate } from '../runtype';
...
+  Part extends boolean,
+  RO extends boolean
+> extends Runtype<RecordStaticType<O, Part, RO>> {
   tag: 'record';
   fields: O;
+  isPartial: Part;
   isReadonly: RO;
 
-  asReadonly(): Record<O, true>;
+  asPartial(): InternalRecord<O, true, RO>;
```
keithlayne, line 27: I threw this in there because it seems like a win for usability.  

The slippery slope includes `asRequired` and `asMutable`.  I'm not sold on those.

src/types/record.ts
```diff
@@ -2,26 +2,52 @@ import { Runtype, Static, create, innerValidate } from '../runtype';
...
+  false,
+  RO
+>;
+
+export type Partial<O extends { [_: string]: Runtype }, RO extends boolean> = InternalRecord<
+  O,
+  true,
+  RO
+>;
+
```
keithlayne, line 42: Prettier hates me.

src/types/record.ts
keithlayne: Fun trick, you could do i.e.

```ts
asPartial(this: { isPartial: false }): InternalRecord<O, true, RO>;
```

and get a type error on when calling this on a `Partial`.

src/reflect.ts
```diff
@@ -13,9 +13,12 @@ export type Reflect =
...
-  | ({ tag: 'record'; fields: { [_: string]: Reflect }; isReadonly: boolean } & Runtype<
-      { readonly [_ in string]: unknown }
-    >)
+  | ({
+      tag: 'record';
+      fields: { [_: string]: Reflect };
+      isPartial: boolean;
+      isReadonly: boolean;
+    } & Runtype<{ readonly [_ in string]: unknown }>)
   | ({ tag: 'partial'; fields: { [_: string]: Reflect } } & Runtype<{ [_ in string]?: unknown }>)
```
pelotom, line 22: The `partial` reflect variant should be removed since it's been subsumed by `record` now, right?

keithlayne: Yep, missed  that, my bad.

keithlayne:
@pelotom How do you feel about the `.asPartial()`?  I figured you didn't hate it at least since you merged it :)

But probably need to document it.  This is sort of a reminder to myself to try to do that.

pelotom:
It seems fine to me. Documentation would be good though!

PR #157, opened 2020-07-16
Export InternamRecord

keithlayne:
This is my fault, since the lib built and tested fine I thought maybe it
was okay to leave out the exports. This change just restores the things
that were formerly exported.

When you install 5.0, you will get compile errors like

```
Exported variable 'Foo' has or is using name 'InternalRecord' from external module ".../node_modules/runtypes/lib/types/record" but cannot be named.
```

if you use `Record`.


keithlayne:
@pelotom I'm sorry, I broke your lib.  :(

pelotom:
No worries, thanks for the quick fix.

Issue #154, opened 2020-07-14
hasKey (via new Discriminated Union code) throws when value being validated is null


NMarrero:
The new Discriminated Union code in 4.3.0 fails when value is null due to a bug in hasKey in https://github.com/pelotom/runtypes/blob/master/src/util.ts. 

Called from https://github.com/pelotom/runtypes/blob/master/src/types/union.ts#L996

If value is null, in ```hasKey``` ```typeof value === 'object'``` is true (thanks Javascript!) so it throws an exception when using the ```in``` operator.

Here's a small example: https://codesandbox.io/s/charming-dubinsky-x2mpy?file=/src/index.ts

pelotom:
Hi NMarrero, thanks for the bug report. Would you be interested in making a PR to fix it?

PR #158, opened 2020-07-24
Passing null to Union(...).guard throws in some cases

keithlayne:
I'm not sure at all why, and the test with the types copied from above
passes, so I'm pretty confused.  I don't have time to debug it right
now, but these cause:

```
TypeError: Cannot use 'in' operator to search for 'kind' in null
```

And here's the backtrace for reference:

```
at Object.hasKey (src/util.ts:4:36)
at Union.runtype_1.create.tag (src/types/union.ts:996:17)
at Object.A._innerValidate (src/runtype.ts:116:12)
at Object.A.validate (src/runtype.ts:118:34)
at Object.guard (src/runtype.ts:139:14)
at Object.<anonymous> (src/types/union.spec.ts:87:33)
```

Am I doing something dumb or missing something obvious?  I can't see a
real difference between the passing case and the failing ones.

### DIFF
--- a/src/types/union.ts
+++ b/src/types/union.ts
@@ -993,6 +993,7 @@ export function Union(...alternatives: Rt[]): any {
               const field = alternative.reflect.fields[fieldName];
               if (
                 field.tag === 'literal' &&
+                value !== null &&
                 hasKey(fieldName, value) &&
                 value[fieldName] === field.value
               ) {
--- a/src/types/union.spec.ts
+++ b/src/types/union.spec.ts
@@ -40,7 +40,7 @@ describe('union', () => {
       expect(Shape.validate({ kind: 'other', size: new Date() })).not.toHaveProperty('key');
     });
 
-    it('hould not pick alternative if the discriminant is not unique', () => {
+    it('should not pick alternative if the discriminant is not unique', () => {
       const Square = Record({ kind: Literal('square'), size: Number });
       const Rectangle = Record({ kind: Literal('rectangle'), width: Number, height: Number });
       const CircularSquare = Record({ kind: Literal('square'), radius: Number });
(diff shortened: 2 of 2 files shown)


keithlayne:
At the point of the error (https://github.com/pelotom/runtypes/blob/master/src/util.ts#L3-L5) `o` is annotated as `{}`, so `null` somehow slipped in there.

I can't believe that this has never come up before, now I'm worried that my PR broke this too.

keithlayne:
Bisecting shows it introduced here: https://github.com/pelotom/runtypes/commit/c90580603e0217f689f11e32db1b2e083fe0748a

src/types/union.ts
```diff
@@ -993,6 +993,7 @@ export function Union(...alternatives: Rt[]): any {
               const field = alternative.reflect.fields[fieldName];
               if (
                 field.tag === 'literal' &&
+                value !== null &&
```
keithlayne, line 996: This "fixes" it, but it's still not clear to me why the other test is not hitting  this.

adriancooney:
> Bisecting shows it introduced here: c905806

For reference @keithlayne, I'm seeing this issue in 4.3.0. Finding it difficult to diagnose the source of the error however.

keithlayne:
@adriancooney - Not sure we're talking about the same thing.  If you look in the tests I added:

```ts
      const DiscriminatedUnion = Union(
        Record({ kind: Literal('foo'), size: Number }),
        Record({ kind: Literal('bar'), foo: String }),
      );

      expect(DiscriminatedUnion.guard(null)).toBe(false);
```

You should be able to pass any value to `guard` without throwing.  `Null` doesn't need to be in the union to trigger this.  It may have the same root cause though, and putting `Null` first may reorder the checks and short-circuit before the exception is thrown.

adriancooney:
Ah right, I see. It was an identical stack trace to yours. I think your final assessment sounds about right to me.

yuhr:
I believe this is the same issue as #154, which was resolved by #205.

PR #159, opened 2020-07-24
More specific runtime guard function signatures

lukebarton:
I wrote this up as an issue, but I figured I'd just go ahead and submit a patch.

I was unsure as to why `check`, and perhaps to a lesser extent `guard` have a type signature of `check(x: any): A` rather than `check(x: A): A` ? At the moment, it kinda seems the wrong way around. The `any` should be on the outside so that the typechecker can infer when we're trying to use the checks/guards wrong, before we get anywhere near runtime.

Let's look at `check` first: `check` even throws an error if `x` is not an `A`. That means there is no valid codepath out of `check` if `x` is not an `A`, therefore couldn't `x: A` on input? If my typechecker knows that `check` expects an `A` and it knows that my `x` is not an `A` (it could) then it could tell me that before I've even run my code. Trivial example: `Number.check("abc")` makes no sense and could be easily typechecked before we get to runtime, yet the type defs happily let that pass, even though we know upfront that this function will not return.

Onto `guard` which is similar, but `guard` returns a boolean. This one is more complicated because a `false` does not break execution like throwing an `Error`, suggesting it might be expected behaviour to pass in something that is not an `A`. However the name `guard` suggests an expectation that `x` _should be_ an `A`, and again, if my typechecker knows that `x` is definitely not an `A`, then it would be nice to be told at typecheck time. Now, if you legitimately use a function called `guard`, knowingly passing in something that your typechecker already knows by type definition, is not an `A`, then you could save yourself an `if` statement there. One use case that springs to mind is if you have something like an `x: Foo | Bar` and you want to `if (Foo.guard(x)) { ... } else if (Bar.guard(x)) { ...}` then you could instead `if (Foo.guard(x as Foo)) { ... } else if (Bar.guard(x as Bar)) { ...}` which would be better than `guard` accepting `any` by type, and losing the ability to typechecker to defend users from situations it can infer are unintended e.g. `String.guard(12)`.

In summary, shouldn't `any` be on the outside of the type definition, not in it?

Tests and typecheck both pass.


lukebarton:
This means that this will fail at typecheck time:

```typescript
const runType = Record({ foo: String, bar: Number });
const value: { qux: number } = { qux: 12};
runType.check(value);
```
However, as you'd expect, this will still pass at typecheck time:
```typescript
const runType = Record({ foo: String, bar: Number });
const value: any = await someApiResponseThatNeedsRuntimeRefinement();
runType.check(value);
```

yuhr:
Hmm... I understand your point, but doing type assertions like `Foo.check(x as Foo)` in almost every `check` looks quite redundant to me. Furthermore, I know this is just a subjective opinion, but the most unpleasant thing is, it shall effectively prohibit passing just `unknown` values to `check` functions without `as any`. I'm sure it would be absolutely weird API...

@pelotom Can I hear your advice about this, if you spare some time?

pelotom:
Hm, I'm not sure I understand. The purpose of `check` is to take an object about which we potentially know nothing and validate that it has the specific static type associated with the runtype. So perhaps the right type is

```ts
check(x: unknown): A;
```

I can't remember if there was some specific reason I used `any` instead of `unknown`; perhaps it's just a legacy from the days before `unknown` existed.

Technically we should be able statically constrain the input to be a _supertype_ of the runtype, but as far as I know there's no good way to express this. For instance if there was a `super` keyword that let you constrain a type to be a supertype of another known type, then this might be the best signature:

```ts
check<B super A>(x: B): A;
```

With conditional types we can sort of approximate this with

```ts
check<B>(x: B): A extends B ? A : never;
```

which states that we know statically that the call will not return successfully if the input type is not a supertype of the runtype in question.

yuhr:
That has to be the right direction I think too. Returning `never` type in invalid codepaths seems to be enough to notify it's statically invalid.

PR #162, opened 2020-08-17
Exact record support

### LINKED ISSUE #41: Add a flag to check/validate/guard which allows enforcing exact match (not just structurally assignable)
Perhaps I'm missing something fundamental here about how this library makes assertions, but the following obviously errors in Typescript:

```
interface ISeries {
  name: string;
  value: number;
}

const series: ISeries = {
  name: 'test',
  value: 10,
  foo: true,
};
```

but this `check()` does not error with runtypes:

```
const Series = Record({
  name: String,
  value: Number,
});

Series.check({
  name: 'test',
  value: 10,
  foo: true
});
```

Why is the additional key `foo` allowed through in this case?

reinismu:
Added `.exact()` for `Record`. It allows to check that incoming object doesn't have additional fields.

Close #41 

First I tried to create `{ strict: true }`, but `validate`, `check` etc. are shared between multiple Runtypes and they don't really need that setting.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -752,10 +752,10 @@ describe('change static type with Constraint', () => {
     | Literal<boolean | number | string>
     | Array<Reflect, false>
     | Array<Reflect, true>
-    | Record<{ [_ in string]: Reflect }, false>
-    | Record<{ [_ in string]: Reflect }, true>
-    | RTPartial<{ [_ in string]: Reflect }, false>
-    | RTPartial<{ [_ in string]: Reflect }, true>
+    | Record<{ [_ in string]: Reflect }, false, false>
+    | Record<{ [_ in string]: Reflect }, true, false>
+    | RTPartial<{ [_ in string]: Reflect }, false, false>
+    | RTPartial<{ [_ in string]: Reflect }, true, false>
     | Tuple2<Reflect, Reflect>
     | Union2<Reflect, Reflect>
     | Intersect2<Reflect, Reflect>
--- a/src/types/record.spec.ts
+++ b/src/types/record.spec.ts
@@ -1,4 +1,4 @@
-import { String, Record, Array } from '..';
+import { String, Record, ExactRecord, Array } from '..';
 
 const AnimalRecord = Record({
   cat: String,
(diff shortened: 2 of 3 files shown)


jraoult:
@reinismu I'm excited about this. Have you heard from @pelotom regarding this PR?

reinismu:
Nop :/ 

I made my own build so I could use this change + Omit
```
"dependencies": {
    "runtypes": "github:reinismu/runtypes#with-pick-omit"
  },
```
If you are interested

bayareacoder:
This doesn't work for me if you have a record with optional properties:
const Test = Record({...}).And({...}).exact();
Seems a common use case to make sure that no extra keys, besides the optional ones, are present.

`Property 'exact' does not exist on type 'Intersect2<Record`

jraoult:
@reinismu cheers, I ended up moving to [zod](https://github.com/colinhacks/zod) anyway. It wasn't too hard after all. I feel this project is sadly losing momentum and zod provide strict check by default out of the box and some other nice goodies API wise.

bayareacoder:
@reinismu You closed this issue but does it work for an object with both required and optional keys, to make sure there are either no required keys missing, or no additional keys, beyond the optional ones, specified? I don't see that case in your tests.
I believe this is a very common use case to eg test an API response.
We use this library and found a way to check for this that works, but is not elegant:

```
const allowedKeys = ['requiredKey1', ..., 'requiredKeyM', 'optionalKey1', ..., 'optionalKeyN'];
const RTtype = Record({
requiredKey1: ...
...
requiredKeyM: ...
})
.And(Partial({
optionalKey1: ...
...
optionalKeyN: ...
})
.And(Dictionary(Unknown, allowedKeys));
```
This will check that:
- all required keys are present and of correct type
- optional keys, if present, are of the correct type
- no additional keys are present.

Basically I don't think an exact spec on any union type works, since it's not the same as the union of the exact types...

reinismu:
@jraoult Thanks! It seems promising  

@bayareacoder From use of it I have noticed few edge cases where it doesn't work. One of them is when trying to use `Or`

yuhr:
Closing in favof of #391.

PR #164, opened 2020-09-04
chore: update TypeScript to 4.0.2

ForbesLindesay:
This will enable features like arbitrary length tuples and functions with arbitrary numbers of parameters.


yuhr:
It's totally pointless to change the name of this package, even if this PR includes any other improvements.

PR #166, opened 2020-09-04
feat: support arbitrary length unions

ForbesLindesay:
I had to separate out "RuntypeBase" which only includes the actual validation methods from "Runtype" which includes all the helper methods. This is because internally we frequently treat Runtype as "covariant" i.e. if I was expecting a `Runtype<string>` it's ok to give me a `Runtype<'hello world'>`. TypeScript historically hasn't cared that much about variance, but over time it is getting strickter and the helper methods are often invariant.

BREAKING CHANGE: the various `Union2`, `Union3`, ... types have all been replaced with a single `Union` type that takes a tuple of Runtype validators.


joual:
What is needed to get this merged?

yuhr:
Please don't include changes that are only relevant to your fork release. This package is not `funtypes`. We can't merge because of it no matter how much the PR includes substantial improvements.

yuhr:
Hi @ForbesLindesay, thank you for your amazing works! May I force push to your PR branches of #165, #166, and #168, rebasing them to the current `master`? And as for #166, I'd like to cherry-pick the essential commit 19e2e865f1bbfbc3bde0743d006cc0aa78c6c2e2. If you have any project depending on these PR branches, let me know so that I won't break them.

PR #170, opened 2020-09-09
feat: ParsedValue type

ForbesLindesay:
This allows you to define a type that has a different representation for parsed vs. serialized data. The `serialize` method is left as optional, which lets you create parse-only codecs for legacy data while always serializing to the latest format


ForbesLindesay:
Sorry, wrong repo again. I can't find any option on GitHub to change the deault.

Issue #111, opened 2019-10-11
Support for stringify and parse methods


maclockard:
I wanted to get your thoughts on a potential feature, if its something that you are open to, I would be more than happy to work on it!

## Overview
Add `stringify`/`parse` methods to Runtype. These methods would work somewhat similar to `JSON.stringify` using the [`replacer` parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) and `JSON.parse` using the [`reviver` parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) to whitelist. However, the runtypes version of `stringify`/`parse` would recurse through the given Runtype and use that to whitelist which properties get serialized/deserialized. 

## Use Case

Currently runtypes supports enforcing object structure aligns with type expectations. This is incredibly useful when sending/receiving data over the wire. It would make runtypes more powerful if it could also be used to enforce that serialization/deserialization works as expected. When using `JSON.stringify` its very easy to accidentally include fields that should not be returned or were not even defined on the runtype. 

If implementing an endpoint that sends an `User` object:

```typescript
import { Record, String } from 'runtypes';

const User = Record({
  userName: String,
  displayName: String,
});

async function getUser(req, res, next): Promise<void> {
  ...
  const rawUser:  = await getUserFromDataBase();
  const typedUser = User.check(rawUser);

  res.body(JSON.stringify(myUser)); // DANGER DANGER could include property 'socialSecurityNumber'
}
``` 

This could be very dangerous if the object returned by `getUserFromDataBase` looked like this:

```typescript
{
  userName: "maclockard",
  displayName: "Mac",
  socialSecurityNumber: 555555, // DANGER DANGER
}
```

since `JSON.stringify` would include the entire object, and I would have thought I was safe given I checked it was a `User` using runtypes. There are multiple different ways to solve this, such as using the [`replacer` parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) or explicitly constructing a new object everytime, however, these become increasingly unwieldy and error prone for complex, large, or deeply nested objects. It also doesn't make sense to change `check` to throwing on unexpected fields since the object is still a valid `User`.

So instead I propose adding a `stringify` method to the Runtype itself, that would handle this whitelisting automatically based on how it was defined.

## Example

```typescript
import { Record, String } from 'runtypes';

const User = Record({
[...]

PR #191, opened 2021-03-15
feat: add `Transform`

### LINKED ISSUE #56: Is converting values in the scope of this library ?
A common use case when validating data from external APIs is to have dates encoded as strings. 
Let's say I get the following data from a JSON API:

```json
{ "date": "2015-10-12 12:00:00" }
```

I would want to represent it with the following Runtype:

```typescript
Record({ date: InstanceOf(Date) })
```

Is there a simple way to go from the first to the second without having to use an intermediate `Record({ date: String })` runtype, and doing the validation ?

I think that having a `convertFrom` method on the `Runtype` interface would help a lot.

```typescript
interface Runtype<A> {
  // ...
  convertFrom<B> : (checkBefore: (x: any) => B) => Runtype<A>;
}
```

The returned `Runtype` would be exactly the same, but the `check` method would be changed to include the additional validation in `checkBefore` before the one from the initial `Runtype`.

I could then do:

```typescript
const Input = Record({ 
  date: InstanceOf(Date).convertFrom(x => {
    const strX = String.check(x);
    const timestamp = Date.parse(strX);
    if(isNan(timestamp)) {
      throw new ValidationError('Invalid date string ' + strX);
    }
    return new Date(timestamp);
 })
});
```

Do you think it could be added to the library (I would be willing to do it)?

yuhr:
BREAKING CHANGE: `check` method of compound runtypes such as `Record` and `Tuple` now returns a new object instead of original one

Resolves #56. It's inspired by https://github.com/pelotom/runtypes/issues/56#issuecomment-757425403. Detailed usage is in the [tests](https://github.com/pelotom/runtypes/pull/191/files#diff-b075331330986a228557d856a506184846e4f748b2b7a5fce7b313fc6d73df45). Above change is needed in order to make `Transform` work with those non-primitive runtypes. Now investigating possible side effects, bugs & design flaws...

Example usage:

```ts
const Input = Record({
  date: String.withTransform(s => {
    const date = new Date(s);
    if (isNaN(date.valueOf())) throw new Error(`Invalid date string "${s}"`);
    else return date;
  }),
});
type Input = Static<typeof Input>;
// { date: Date; }

Input.check({ date: 'crap' });
// ValidationError: Failed transform: Error: Invalid date string "crap" in date

const input: Input = Input.check({ date: '1990-01-01' });
console.log(input.date.toUTCString());
// Mon, 01 Jan 1990 00:00:00 GMT
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,47 @@ const B = Guard(Buffer.isBuffer);
 type T = Static<typeof B>; // T will have type of `Buffer`
 ```
 
+## Transforms
+
+Sometimes we want to convert values along with validation, for example:
+
+- Populate properties of more descriptive object with a shorthand value
+- Deserialize incoming data into hydrated object, e.g. timestamp `number` to `Date`
+- Maintain backward compatibility upgrading old input into new representation
+
+We have a Swiss Army knife for such situations, called `Transform`. Usually you
+want to use it through `withTransform` method of a `Runtype`:
+
+```ts
+const ParsedInt = String.withTransform(s => {
+  const parsed = parseInt(s);
+  if (isNaN(parsed)) throw new Error('Parse failed');
+  else return parsed;
+});
+```
+
+Errors thrown by transformer functions are gracefully caught by the underlying
+`Runtype` and treated as validation errors like usual:
+
+```ts
+if (ParsedInt.guard('crap')) return 'Never reach here';
+else return 'Always reach here';
+
+ParsedInt.check('crap');
+// ValidationError: Failed transform: Error: Parse failed
+```
+
+`Static` should work as expected:
+
+```ts
+type ParsedInt = Static<typeof ParsedInt>;
+// Above type is inferred as: `number`
+```
+
+Maybe you've noticed that `Transform` actually covers the same functionalities
+of `Constraint` and `Guard`, but it's not recommended using `Transform`s without
+any transformation of values, simply because it makes your code look nonsense.
+
 ## Function contracts
 
 Runtypes along with constraint checking are a natural fit for enforcing function
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -29,6 +29,7 @@ import {
   InstanceOf,
   Brand,
   Guard,
+  Transform,
 } from './index';
 
 import { Constructor } from './types/instanceof';
@@ -173,6 +174,16 @@ const runtypes = {
     .asReadonly()
     .And(RTPartial({ bar: String }).asReadonly()),
   EmptyTuple: Tuple(),
+  ParseInt: String.withTransform(s => {
+    const parsed = parseInt(s);
+    if (isNaN(parsed)) throw new Error('Input is invalid');
+    else return parsed;
+  }),
+  AbsoluteNumber: Number.withTransform(n => {
+    const abs = Math.abs(n);
+    if (isNaN(abs)) throw new Error('Input is invalid');
+    else return abs;
+  }).withBrand('AbsoluteNumber'),
 };
 
 type RuntypeName = keyof typeof runtypes;
@@ -188,11 +199,19 @@ const testValues: { value: unknown; passes: RuntypeName[] }[] = [
   { value: null, passes: ['Null', 'Void'] },
   { value: true, passes: ['Boolean', 'true'] },
   { value: false, passes: ['Boolean', 'false'] },
-  { value: 3, passes: ['Number', 'brandedNumber', 3, 'union1'] },
+  { value: 3, passes: ['Number', 'brandedNumber', 3, 'union1', 'AbsoluteNumber'] },
   {
     value: 42,
-    passes: ['Number', 'brandedNumber', 42, 'MoreThanThree', 'MoreThanThreeWithMessage'],
+    passes: [
+      'Number',
+      'brandedNumber',
+      42,
+      'MoreThanThree',
+      'MoreThanThreeWithMessage',
+      'AbsoluteNumber',
+    ],
   },
+  { value: '42', passes: ['String', 'union1', 'ParseInt'] },
   { value: 'hello world', passes: ['String', 'hello world', 'union1'] },
   { value: [Symbol('0'), Symbol(42), Symbol()], passes: ['symbolArray'] },
   { value: Symbol.for('runtypes'), passes: ['Sym'] },
@@ -716,6 +735,13 @@ describe('reflection', () => {
     expectLiteralField(C, 'tag', 'brand');
     expectLiteralField(C.entity, 'tag', 'number');
   });
+
+  it('transform', () => {
+    const C = Number.withTransform(Math.abs, { name: 'AbsoluteNumber' });
+    expectLiteralField(C, 'tag', 'transform');
+    expectLiteralField(C.underlying, 'tag', 'number');
+    expectLiteralField(C, 'name', 'AbsoluteNumber');
+  });
 });
 
 describe('change static type with Constraint', () => {
@@ -762,7 +788,8 @@ describe('change static type with Constraint', () => {
     | Function
     | Constraint<Reflect, any, any>
     | InstanceOf<Constructor<never>>
-    | Brand<string, Reflect>,
+    | Brand<string, Reflect>
+    | Transform<Reflect, unknown>,
 ): Reflect => {
   const check = <A>(X: Runtype<A>): A => X.check({});
   switch (X.tag) {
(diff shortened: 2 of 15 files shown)


yuhr:
https://github.com/ForbesLindesay/funtypes/issues/16#issuecomment-690852229

> A side effect of this change is that we are also effectively deep cloning everything,

This is correct for us too. 

> which takes care of pelotom#132 and pelotom#24 (remove unknown keys from validated object).

I think this is still remaining to be considered in some point of view, that is, we could still include unknown properties in the validated object, but yes, it would be better to remove in favor of these issues. This way, we shall state that **`Record` and `Partial` will always strip unknown properties** in the validated object, so if one wants to include them in the output, explicit runtypes are needed for them, for example intersecting `Dictionary` or somehow.

Additionally, this will consequently handle "exact" type matching #41. It feels like we're somewhat on the right track further than #162, as it won't work with `Intersect` or `Union`.

> It may be worth looking into supporting a fast path that doesn't clone objects if none of the validations change anything.

I don't believe this would improve performance or memory consumption in any terms, because it requires us to introduce additional strict-equality check at every single runtype validation.

yuhr:
Note that `Transform` won't work with decorators, as decorators currently don't support any transform of static types of their targets such as parameters. https://github.com/microsoft/TypeScript/issues/4881

yuhr:
The side effect that it removes unknown properties from validated records may also address the problem described in #111:

> ```ts
> import { Record, String } from 'runtypes';
> 
> const User = Record({
>   userName: String,
>   displayName: String,
> });
> 
> async function getUser(req, res, next): Promise<void> {
>   ...
>   const rawUser:  = await getUserFromDataBase();
>   const typedUser = User.check(rawUser);
> 
>   res.body(JSON.stringify(myUser)); // DANGER DANGER could include property 'socialSecurityNumber'
> }
> ```
> 
> 
> This could be very dangerous if the object returned by getUserFromDataBase looked like this:
> ```ts
> {
>   userName: "maclockard",
>   displayName: "Mac",
>   socialSecurityNumber: 555555, // DANGER DANGER
> }
> ```

Obviously this will already be resolved by this PR.

As I have written [here](https://github.com/pelotom/runtypes/issues/56#issuecomment-799833420), once we get `Transform`, the "stringify" feature proposed by #111 is easily created by having another runtype dedicated to serialization.

yuhr:
Well, I understand in some cases it's not "easily created", especially for complex `Record`s. In those cases we might think of having some convenient way to generate a new runtype derived from existing one. This conversation is out of scope of this PR, since this PR is supposed to provide a basic functionality to perform data transformation.

yuhr:
Oh this will break the semantics of `guard` and `assert` now, that needs a solution definitely.

peterjwest:
Looking forward to this feature, will be super useful!

peterjwest:
It is a shame the non-mutating behaviour will be lost though, since it's so low overhead.

yuhr:
@peterjwest Sorry, my wording was improper. I said misleading terms such as "strip", "remove" and "side effect", but by these words I did mean just "not present" in the returned object (and its static type) from ".check()". That's why the current implementation makes `guard` and `assert` nonsense, i.e. `Transform` can change the `Static` type of a runtype, but `guard` and `assert` does not return the transformed value and just assure that *the argument's type is the transformed one* (Of course this behaviour is invalid, and needs to be resolved). I don't want to introduce any form of mutating behaviour.

peterjwest:
I think I understand what you're saying, and that sounds great! But in regard to your question, I have absolutely no idea.

milankinen:
Any news about this? This feature would be very important to our team if it gets ready (especially if extra object keys will be dropped). So far we've used this kind of utility function but it feels a bit hacky and we'd rather implement this as real real feature to the core library:
```ts
const withoutExtraKeys = <T>(type: rt.Runtype<any>, value: any): T => {
  const reflect = type.reflect
  switch (reflect.tag) {
    case 'record':
      return Object.entries(reflect.fields).reduce((result, [propertyKey, propertyType]) => {
        if (propertyType.tag !== 'optional' || propertyKey in value) {
          result[propertyKey] = withoutExtraKeys(propertyType, value[propertyKey])
        }
        return result
      }, {} as T)
    case 'dictionary':
      const valueType = reflect.value
      return Object.entries(value).reduce((res, [key, value]) => {
        res[key] = withoutExtraKeys(valueType, value)
        return res
      }, {} as T)
    case 'array':
      const elementType = reflect.element
      return value.map((x) => withoutExtraKeys(elementType, x))
    case 'optional':
      return value === undefined ? value : withoutExtraKeys(reflect.underlying, value)
    case 'union':
      for (const altType of reflect.alternatives) {
        if (altType.validate(value).success) {
          return withoutExtraKeys(altType, value)
        }
      }
      return value
    case 'intersect':
      if (reflect.intersectees.length === 0) {
        return value
      } else {
        return reflect.intersectees.reduce((result, intersecteeType) => {
          const intersecteeValue = withoutExtraKeys(intersecteeType, value)
          if (_.isPlainObject(result) && _.isPlainObject(intersecteeValue)) {
            Object.assign(result, intersecteeValue)
            return result
          } else {
            return intersecteeValue
          }
        }, {} as any)
      }
    default:
      return value
  }
}

export const checkNoExtraKeys = <T>(type: rt.Runtype<T>, value: unknown): T => {
  type.check(value)
  return withoutExtraKeys<T>(type, value)
}```

I can also help with this PR if it helps to get it closer to being merged!

**Comment by peterjwest** [comment]
Here's my version of the same thing.
It's not as feature complete as yours, but mine returns the original object unless it would be mutated.

```ts
import { mapValues, pickBy } from 'lodash';
import * as runtypes from 'runtypes';

export default function restrict<Data, RunType extends runtypes.Runtype<Data>>(RunType: RunType, data: Data) {
  return restrictInternal(RunType, data) || data;
}

function restrictInternal<Data, RunType extends runtypes.Runtype<Data>>(
  RunType: RunType, data: Data,
): Data | undefined {
  const reflected = RunType.reflect;
  if (reflected.tag === 'record') {
    let mutated = false;

    const newData: Data = mapValues(pickBy(data as any, (item, key) => {
      const hasKey = key in reflected.fields;
      if (!hasKey) mutated = true;
      return hasKey;
    }), (item, key) => {
      const newItem = restrictInternal(reflected.fields[key], item);
      if (newItem !== undefined) mutated = true;
      return newItem || item;
    });

    if (mutated) return newData;
  }
  if (reflected.tag === 'array') {
    let mutated = false;

    const newData: Data = (data as any).map((item: any) => {
      const newItem = restrictInternal(reflected.element, item);
      if (newItem !== undefined) mutated = true;
      return newItem || item;
    });

    if (mutated) return newData;
  }
  if (reflected.tag === 'dictionary') {
    let mutated = false;

    const newData: Data = mapValues(data as any, (item) => {
      const newItem = restrictInternal(reflected.value, item);
      if (newItem !== undefined) mutated = true;
      return newItem || item;
    });

    if (mutated) return newData;
  }

  if (reflected.tag === 'union') {
    const alternative = reflected.alternatives.find((alternative) => {
      try {
        alternative.check(data);
        return true;
      } catch {
        return false;
      }
    });
    return restrictInternal(alternative as any, data);
  }

  if (reflected.tag === 'optional') {
    return restrictInternal(reflected.underlying as any, data);
  }

  return undefined;
}

```

jgrigg:
If we are transforming in the direction of Json => TS is there then also appetite for transformations in the opposite direction i.e TS => Json? Presumably we'd need a new function like `encode()` à la [io-ts](https://github.com/gcanti/io-ts/blob/master/index.md#the-idea)

WilliamABradley:
@yuhr design-wise, shouldn't withTransform be inverted? Or am I thinking more on the lines of a withConverter function.

Ideally my thought it that I have:

```ts
export const DateType = rt.InstanceOf(Date).withConverter((val: unknown) => {
    if (typeof value === 'string') {
    value = new Date(DateString.check(value));
  }
  if (DateTime.isDateTime(value)) {
    value = value.toJSDate();
  }
  return value;
});
export type DateType = rt.Static<typeof DateType>;
```

That way it can deal with any kind of unknown value, and try and turn it into the Date Type, and if it fails, throw a validation error.

To represent this with your withTransform function, do I use a union of possible input types?

E.g.

```ts
export const DateType = rt.Union(rt.InstanceOf(Date), rt.InstanceOf(DateTime), rt.String).withTransform((val: Date | DateTime | string) => {
    if (typeof value === 'string') {
    value = new Date(DateString.check(value));
  }
  if (DateTime.isDateTime(value)) {
    value = value.toJSDate();
  }
  return value;
});
export type DateType = rt.Static<typeof DateType>;
```

yuhr:
Closing as too slate; I'll file another PR for this.

PR #193, opened 2021-03-21
Feature: generic dictionary key

### LINKED ISSUE #192: Better key type of `Dictionary`
As noticed in #79, #143, and #160, we'd better implement the type of keys of `Dictionary` more generic way. Current implementation is taking a type as a string, but using unions and enums as the type of key of `Dictionary` is pretty common, so we must handle this. Possible API usage should look like:

```ts
Dictionary(RuntypeForValue, RuntypeForKey)
```

This way, we would deprecate string input of the second parameter, and check the `RuntypeForKey` to be valid as a key type in current TS.

yuhr:
Resolves #192.


KingDarBoja:
A great addition, awesome! No issues so far with this new feature 🚀

PR #198, opened 2021-03-24
Add workflow to generate and push docs to Wiki

### LINKED ISSUE #197: Generate API reference from the codebase
It may be good to have a full API reference to provide an exhaustive list of available APIs, as oftenly people missing APIs undocumented in README.md. I think GitHub Pages would fit, as we want to auto-generate it somehow (e.g. using [TypeDoc](https://github.com/TypeStrong/typedoc)) from the codebase. Or we may utilize GitHub Wiki as well, by using [typescript-documentation](https://github.com/mucsi96/typescript-documentation).

yuhr:
Resolves #197. I'm using [`typedoc`](https://www.npmjs.com/package/typedoc) and [`typedoc-plugin-markdown`](https://www.npmjs.com/package/typedoc-plugin-markdown) in order to generate docs as Markdown files. You can see how it looks here: https://github.com/yuhr/runtypes/wiki

Maybe we could use GitHub Pages or GitBook, but I'm not sure what advantages they would have actually. GitHub Wiki seems to have a great integration here, as it automatically populates links to the source code i.e. "Defined in" (you can see for example [here](https://github.com/yuhr/runtypes/wiki/array#properties-1), this is not a thing handled by `typedoc`).


yuhr:
Oops, we should use this: https://github.com/tgreyuk/typedoc-plugin-markdown/tree/master/packages/typedoc-github-wiki-theme

arnulfojr:
Any other option?

Maybe we can focus more on updating the README and publish it as a website like in zod? I'm not sure if we really need to have API docs in runtypes. Based on my experience the pain points around the library are around "examples", "how-to's", or "what can we do with this lib?" and updating the README would be good enough?

e.g., include Dictionary in the readme

Issue #199, opened 2021-03-24
Support symbols as `Record` keys


yuhr:
Currently (at least `5.1.0`), [`Record` uses `for..in` syntax](https://github.com/pelotom/runtypes/blob/v5.1.0/src/types/record.ts#L59) to loop over fields, this way properties with keys being symbols are just ignored. We should handle this, ~~but I wonder if this is a kind of breaking change or just a bugfix...~~ Anyway this will change current behaviour somehow, so it must be treated as a breaking change.

yuhr:
By the way, the related TS feature will be out there about two months away: https://github.com/microsoft/TypeScript/issues/42762, https://github.com/microsoft/TypeScript/pull/26797. Maybe we should wait for it?

PR #204, opened 2021-03-28
Feature: support symbol-keyed properties of object validation

yuhr:
BREAKING CHANGE: `Dictionary` now does not look over non-enumerable properties of tested objects, and `Record` now does not recognize inherited properties of passed schema

Resolves: #199.


yuhr:
`Dictionary(String, Number)` is now equivalent to `Array(String)`.

Issue #39, opened 2018-02-12
Provide error meta data


zaaack:
It would be great to provide error meta data, something like this:

```js
let error = {
 success: false,
 message: '...',
 path: ['key1', 'key2', ...],
}
```

This can be useful for form validation because we validate the whole form with an object type, but need to show the message in each field.

pelotom:
Yeah, giving more structured error information is definitely a goal. Thanks for making the issue.

riston:
Regarding nested Record types if one of the nested child types is `null` or `undefined` you get really hard to track error `Expected a defined non-null value but was ${typeof x}` https://github.com/pelotom/runtypes/blob/b4f30656cbeb9ec538471bbaa14fc405813a9d54/src/types/record.ts#L16. 
My proposal is to add some sort of statical name(optional) also which could be used for displaying errors:
`const Person = Record({ name: String, age: Number}, 'Person'?);`

erkiesken:
PR #51 partly addresses this with more descriptive error messages with full failing check path in message.

Ciantic:
One way could be to have `NamedRecord('Person', { name: String, age: Number})`, maybe the name could be used elsewhere, e.g. in SQL library that automatically understand the `person` named field as person record or a `personId` as a foreign key.

pelotom:
@Ciantic that's an interesting idea... or perhaps just a general method that let you name any runtype:

```ts
const Foo = Union(
  // ...
).named('Foo');
```

elderapo:
I think it would be a great idea to add `Record.getErrors` and return array of errors ex:
```
[
	{ key: "name", expectedType: "string" receivedType: "number", value: 33 },
	{ key: "age", expectedType: "number" receivedType: "string", value: "Tom" },
]
```
so user will be able to create custom error messages or even create informative error messages in multiple languages.

gDelgado14:
Regarding https://github.com/pelotom/runtypes/pull/51 ... Is the intended behaviour of the MR to describe all the keys which do not adhere to a parser? 

See the example below, where the error is only specifying that `name` is wrong. But `age` is also wrong.

```typescript
const { Number: rtNumber, Record, String: rtString } = require('runtypes')

const r = Record({
  name: rtString,
  age: rtNumber
})

// both fields are wrong
// but runtypes only says that name is wrong
console.log(r.validate({ name: 12, age: 'gio' }))

/*
  { success: false,
  message: 'Expected string, but was number',
  key: 'name' }
*/
[...]

PR #209, opened 2021-03-29
Feature: provide descriptive error messages

yuhr:
BREAKING CHANGE: validation result now can contain a failcode and a detailed object enumerating where the validation failed exactly

For example:

```ts
const Square = Record({ kind: Literal("square"), size: Number });
const Rectangle = Record({ kind: Literal("rectangle"), width: Number, height: Number });
const Circle = Record({ kind: Literal("circle"), radius: Number });
const Shape = Union(Square, Rectangle, Circle);
console.log(Shape.validate({ kind: "rectangle", size: new Date() }));
```

Above code should output:

```ts
{
  success: false,
  code: 'CONTENT_INCORRECT',
  message: 'Expected { kind: "rectangle"; width: number; height: number; }, but was incompatible',
  details: {
    width: 'Expected number, but was missing',
    height: 'Expected number, but was missing'
  }
}
```

And if:

```ts
try {
  Shape.check({ kind: "rectangle", size: new Date() });
} catch (e) {
  console.log(e);
}
```

Then it emits:

```ts
ValidationError: Expected { kind: "rectangle"; width: number; height: number; }, but was incompatible
    at new ValidationError
    ...
    <stack trace>
    ... {
  code: 'CONTENT_INCORRECT',
  details: {
    width: 'Expected number, but was missing',
    height: 'Expected number, but was missing'
  }
}
```

Following is the full list of failcodes, excerpted form the code:

```ts

/**
 * A predefined error code indicating what type of failure has occured.
 */
export type Failcode = typeof Failcode[keyof typeof Failcode];
export const Failcode = {
  /** The type of the received primitive value is incompatible with expected one. */
  TYPE_INCORRECT: 'TYPE_INCORRECT',
  /** The received primitive value is incorrect. */
  VALUE_INCORRECT: 'VALUE_INCORRECT',
  /** The key of the property is incorrect. */
  KEY_INCORRECT: 'KEY_INCORRECT',
  /** One or more elements or properties of the received object are incorrect. */
  CONTENT_INCORRECT: 'CONTENT_INCORRECT',
  /** One or more arguments passed to the function is incorrect. */
  ARGUMENT_INCORRECT: 'ARGUMENT_INCORRECT',
  /** The value returned by the function is incorrect. */
  RETURN_INCORRECT: 'RETURN_INCORRECT',
  /** The received value does not fulfill the constraint. */
  CONSTRAINT_FAILED: 'CONSTRAINT_FAILED',
  /** The property must be present but missing. */
  PROPERTY_MISSING: 'PROPERTY_MISSING',
  /** The property must not be present but present. */
  PROPERTY_PRESENT: 'PROPERTY_PRESENT',
  /** The value must not be present but present. */
  NOTHING_EXPECTED: 'NOTHING_EXPECTED',
} as const;
```


yuhr:
We must dig into this point: https://github.com/pelotom/runtypes/issues/39#issuecomment-401983070. Maybe we'd introduce "error code" or similar concept?

yuhr:
With this PR, the existing APIs like `check` and `validate` would suffice normal use case of validation i.e. for logging or debugging. As for further use cases such as displaying multilingual errors to users, now internal implementation of enumerating errors is already compatible for returning more detailed information such as failcodes per property, so we would implement such feature as a separate API for example `inspect` in some upcoming minor versions.

oviniciuslara:
@yuhr 
Hey, is there a way to get detailed information from check's possible exception? I don't want to use validate because it can't cast to the type. 

See how I'm getting the info right now:
```
try {
    const user: User = User.check(data);
} catch (error) {
    console.log(User.validate(data));
}
```

PR #218, opened 2021-04-07
Feature: add generic `properties` accessor for complex runtypes

### LINKED ISSUE #217: Add generic `properties` accessor for complex runtypes
I've come up with this idea during investigating #216. Currently, accessing inner runtypes of complex ones such as `Record`, `Tuple`, etc. requires you to write cumbersome code like following:

```ts
runtype.reflect.tag === 'record' ? runtype.reflect.fields : runtype.reflect.tag === 'tuple' ? runtype.reflect.components : ...
```

But I want to write just this instead:

```ts
runtype.properties
```

This should be possible because JS and TS maintain interoperability between objects and arrays, i.e. arrays can be treated as objects with keys being `string` of decimal numbers. If implemented properly, this should also work with `Union` and `Intersect`, because TS already does it for static typing, for example:

```ts
type U = { foo: true } | { foo: false };
let u!: U;
u.foo; // inferred as `boolean`

type I = { foo: boolean } & { foo: boolean; bar: string };
let i!: I;
i.foo; // inferred as `boolean`
i.bar; // inferred as `string`
```

So you should be able to do that on runtypes:

```ts
const U = Union(Record({ foo: Literal(true) }), Record({ foo: Literal(false) }));
u.properties.foo; // should be `Union(Literal(true), Literal(false))` or just `Boolean`

const I = Intersect(Record({ foo: Boolean }), Record({ foo: Boolean; bar: String }));
i.properties.foo; // should be `Intersect(Boolean, Boolean)` or just `Boolean`
i.properties.bar; // should be `String`
```

yuhr:
Closes #217. ~~This is still a very draft, but it would work. Test cases are not yet sufficient almost certainly.~~

### DIFF
--- a/src/types/array.ts
+++ b/src/types/array.ts
@@ -1,18 +1,22 @@
 import { Reflect } from '../reflect';
 import { Details, Result } from '../result';
 import { Runtype, RuntypeBase, Static, create, innerValidate } from '../runtype';
-import { enumerableKeysOf, FAILURE, SUCCESS } from '../util';
+import { enumerableKeysOf, FAILURE, isNumberLikeString, SUCCESS } from '../util';
+import { Never } from './never';
 
 type ArrayStaticType<E extends RuntypeBase, RO extends boolean> = RO extends true
   ? ReadonlyArray<Static<E>>
   : Static<E>[];
 
-interface Arr<E extends RuntypeBase, RO extends boolean> extends Runtype<ArrayStaticType<E, RO>> {
+interface Arr<E extends RuntypeBase, RO extends boolean = boolean>
+  extends Runtype<ArrayStaticType<E, RO>> {
   tag: 'array';
   element: E;
   isReadonly: RO;
 
   asReadonly(): Arr<E, true>;
+
+  readonly properties: { [_: number]: E };
 }
 
 /**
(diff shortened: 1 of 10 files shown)


src/runtype.ts
```diff
@@ -111,24 +111,24 @@ export function create<A extends Runtype>(
...
+    guard: guard,
+    Or: Or,
+    And: And,
+    optional: optional,
+    withConstraint: withConstraint,
+    withGuard: withGuard,
+    withBrand: withBrand,
+    reflect: A,
+    toString: () => `Runtype<${show(A)}>`,
+  });
```
yuhr, line 131: Revert this change, as it is not relevant here.

src/types/array.ts
```diff
@@ -22,7 +26,13 @@ function InternalArr<E extends Runtype, RO extends boolean>(
   element: E,
   isReadonly: RO,
 ): Arr<E, RO> {
-  const self = ({ tag: 'array', isReadonly, element } as unknown) as Reflect;
+  const properties = new Proxy(
+    {},
+    {
+      get: (_, key) => (typeof key === 'string' && !isNaN(+key) ? element : Never),
```
yuhr, line 32: This guard is incorrect. Use `isNumberLikeString` instead.

yuhr:
Closing as stale base branch.

Issue #212, opened 2021-03-31
TypeScript throwing a "Type instantiation is excessively deep and possibly infinite" error


markusylisiurunen:
Hi!

While trying to update `runtypes` from version `5.1.0` to `5.2.0`, we encountered an issue with a fairly straightforward type definition and TypeScript throwing a `Type instantiation is excessively deep and possibly infinite` compilation error. The issue seems to only happen when first compiling a TypeScript package and then using the compiled package in a TypeScript project (so not if the type is used within the same project before compiling). We are using Yarn Workspaces so I created a minimal example where this issue happens.

You can find the example here: [https://github.com/markusylisiurunen/runtypes-excessive-depth-error](https://github.com/markusylisiurunen/runtypes-excessive-depth-error)

Here's a quick way to test:

```sh
git clone git@github.com:markusylisiurunen/runtypes-excessive-depth-error.git
cd runtypes-excessive-depth-error

yarn && cd packages/types && yarn build && cd ../client && yarn build
```

I think the example repository is the simplest way to describe the issue. If you need more details, let me know!

markusylisiurunen:
I tried to do some digging but couldn't find a fix (at least yet). However, I managed to pinpoint the issue down to the changes made to the `Record` type between the versions `5.1.0` and `5.2.0`. It seems to have something to do with the conditional type for `FilterOptionalKeys` and `FilterRequiredKeys`.

So, this is what the types look at v5.2.0

```ts
type FilterOptionalKeys<T> = Exclude<
  {
    [K in keyof T]: T[K] extends Optional<any> ? K : never;
  }[keyof T],
  undefined
>;

type FilterRequiredKeys<T> = Exclude<
  {
    [K in keyof T]: T[K] extends Optional<any> ? never : K;
  }[keyof T],
  undefined
>;
```

and here they are after my changes

```ts
type FilterOptionalKeys<T> = Exclude<
  {
    [K in keyof T]: K;
  }[keyof T],
  undefined
>;

type FilterRequiredKeys<T> = Exclude<
  {
    [K in keyof T]: K;
  }[keyof T],
  undefined
>;
```

I know that the changes break the optional and that this is not a feasible solution. However, by removing the conditional type on `Optional<any>` the errors went away from my toy example and for our real project as well. So, definitely seems to have something to do with this.

yuhr:
[...]

PR #224, opened 2021-04-16
Refactor: widen type parameter boundary to `RuntypeBase`

yuhr:
Related to #212. This will simplify type instanciation substantially.


yuhr:
Oops, wrong branch.

PR #227, opened 2021-04-19
Handle objects with null prototypes

72636c:
It's currently possible to crash e.g. `Runtype.validate` by passing in an object with a null prototype. This may seem like an odd thing to do, but it's employed by some popular packages like Apollo Server and Koa, and can be observed by trying to validate some Koa query params.

```
TypeError: object?.propertyIsEnumerable is not a function
    at ./node_modules/runtypes/lib/util.js:37:126
    at Array.filter (<anonymous>)
    at Object.enumerableKeysOf (./node_modules/runtypes/lib/util.js:37:35)
    at ./node_modules/runtypes/lib/types/record.js:39:122
    at Object.A._innerValidate (./node_modules/runtypes/lib/runtype.js:14:16)
    at Object.A.validate (./node_modules/runtypes/lib/runtype.js:16:46)
```

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -514,6 +514,18 @@ describe('check errors', () => {
     );
   });
 
+  it('record for null prototype', () =>
+    assertAccepts(
+      Object.assign(Object.create(null), {
+        name: 'Jack',
+        age: 10,
+      }),
+      Record({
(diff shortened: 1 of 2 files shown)


yuhr:
Looks good, thanks~

Issue #232, opened 2021-04-20
CI failing in v6.2.0


yuhr:
https://travis-ci.org/github/pelotom/runtypes/builds/767765982

On my local machine, `yarn test` didn't report these errors, but my VSCode throws them as red highlights if I open index.spec.ts in a tab... It seems that there's a configuration mismatch somewhere.

I have no clue why this happens. These errors strangely disappear if I revert #230.

BTW, the error shows some bloated inferred types such as `Intersect<[Intersect<[Intersect<[Intersect<[Union<[Union<[Runtype<Person>, any]>, any]>, any]>, any]>, any]>, any]>`. Maybe this is the very cause of #212?

yuhr:
Not reproducible anymore.

PR #233, opened 2021-04-20
Revert "Remove workaround for resolved TypeScript issue (#230)"

yuhr:
This reverts commit cb8809d3bb5c590ab76cee983ecf7eb66ce36087.

See #232. Apparently #230 broke `Lazy` runtype, so we should revert it until we have a solution for this.


NeverLiez (commented):
Jö

yuhr:
@NeverLiez ???

Issue #248, opened 2021-05-04
Handle `Never` properties in `Record`


yuhr:
https://github.com/pelotom/runtypes/issues/247#issuecomment-831716139:

> I thought it could be achieved by this:
> 
> ```ts
> Union(Record({ a: Unknown, b: Never }), Record({ a: Never, b: Unknown }))
> ```
> 
> But it doesn't work. I believe this is a bug (`Never` property is not working expectedly), and will be fixed soon.

While working on this, I've been also surprised by the fact that TS doesn't handle `never` properties in the way I expected:

```ts
const x: { foo: never } = {};
// Property 'foo' is missing in type '{}' but required in type '{ foo: never; }'.(2741)
```

Consequently, `Never` properties in `Record`s do not work expectedly in the static types:

```ts
const R = Record({ foo: Never });
type R = Static<typeof R>;
const r: R = {}; // This errors at compile-time anyway, while we can make the validation below passing
expect(R.guard(r)).toBe(true);
```

I filed an issue upstream https://github.com/microsoft/TypeScript/issues/43954. I'm going to wait until the TS team tells us how `never` should work, and we would conform to the behavior there.

yuhr:
The question I filed upstream was answered as "Working as Intended", I was just misunderstanding what `never` is meant to be:

> A type with a property of type `never` has zero legal values; `{ }` is not one of them.

We'd follow this direction, making the behavior of `Never` like this:

```ts
const R = Record({ foo: Never })
type R = Static<typeof R>;
const r: R = {}; // This errors at compile-time expectedly
expect(R.guard(r)).toBe(false); // Always fails
```

If one wants to assure a property to be absent, optionalization is needed:

```ts
const S = Record({ foo: Optional(Never) })
type S = Static<typeof S>;
const s: S = {};
expect(S.guard(s)).toBe(true);
```

It's been already working, so we don't need to take any particular action about this. Though users have to wait https://github.com/microsoft/TypeScript/pull/43947 to be merged in order to work with static types properly i.e. it incorrectly allows `undefined` values for now:

```ts
const x: S = { foo: undefined }; // No errors without `--strictOptionalProperties`
```

RohanHart:
I had to search for this issue. Perhaps expose a simple wrapper to make this easier to apply in future
```
export const Missing = Never.optional();
```

PR #249, opened 2021-05-04
Fix: handle `Never` properties in `Record`

yuhr:
This is an ahead-of-time implementation of #248. May not be merged depending on upstream decision.


yuhr:
Closing. See #248.

Issue #275, opened 2021-08-05
More helpful `union` error messages


johngeorgewright:
I recently received a confusing error message from union validation. Take the following as an example... a simple union of 2 objects:

```typescript
import * as t from 'runtypes'

const S = t.String.withConstraint((x) => x === 'bar' || 'Should be bar')

const R = t.Record({ s: S }).Or(t.Record({ t: S }))

try {
  R.check({ s: 'rab' })
} catch (error) {
  console.error(error)
}
```

This produces:

```
ValidationError: Expected { s: string; } | { t: string; }, but was object
    at ...
    <stack traces>
    at ... {
  code: 'TYPE_INCORRECT'
}
```

The message `Expected { s: string; } | { t: string; }, but was object` isn't all that helpful. Can this message be improved somehow?

guyellis:
@johngeorgewright - love what you're doing with this branch. Is there an ETA on merging and publishing this?

johngeorgewright:
@guyellis, as the PR contains a breaking change the runtypes team will release in the next major version. They'll have a better idea when that'll be.

If you're desperate to use this feature, you can swap out the npm dependency with a git dependency: 

```npm i github:johngeorgewright/runtypes#feat/union-error-details```

PR #276, opened 2021-08-05
feat(union): add details to union errors

johngeorgewright:
This is a simple workaround for #275 providing details on all validation errors.

### DIFF
--- a/src/types/union.ts
+++ b/src/types/union.ts
@@ -2,6 +2,7 @@ import { Runtype, RuntypeBase, Static, create, innerValidate } from '../runtype'
 import { LiteralBase } from './literal';
 import { FAILURE, hasKey, SUCCESS } from '../util';
 import { Reflect } from '../reflect';
+import { Details } from '../result';
 
 export interface Union<A extends readonly [RuntypeBase, ...RuntypeBase[]]>
   extends Runtype<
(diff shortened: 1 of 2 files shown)


yuhr:
Thanks I really appreciate this, but changing the failcode is a breaking change. Let me postpone merging this until the next major version.

yuhr:
@johngeorgewright Thanks rebasing this, but basically I don't want to require PR authors to do it by themselves. I'm working on other tasks, and I believe you will feel it tiresome if you need rebasing/merging upstream every time any update occurs. Please leave it to me 😉

src/types/union.ts
```diff
@@ -70,10 +67,18 @@ export function Union<T extends readonly [RuntypeBase, ...RuntypeBase[]]>(
...
-    return FAILURE.TYPE_INCORRECT(self, value);
+    function validate() {
+      const details: Array<Details | string> = [];
+      for (const alternative of alternatives) {
+        const validation = innerValidate(alternative, value, visited);
+        if (validation.success) return SUCCESS(value);
+        else if (validation.details) details.push(validation.details);
+        else details.push(validation.message);
+      }
+      return FAILURE.CONTENT_INCORRECT(self, details);
```
zaverden, line 80: Hi. What if we split these changes into 2 PRs?

PR1: return `code: TYPE_INCORRECT` but with details
PR2: change `code` to `CONTENT_INCORRECT`

In this case PR1 does not have breaking changed, it can be merged in minor release

```suggestion
      const fail = FAILURE.CONTENT_INCORRECT(self, details);
      // TODO: remove this code hack in next major release
      fail.code = Failcode.TYPE_INCORRECT;
      return fail;
```

zaverden: @yuhr what do you think?
would you merge this PR if it preserves `code: TYPE_INCORRECT` but still adding details?

yuhr: Hm, sounds neat but how we should format the error message might differ between `TYPE_INCORRECT` and `CONTENT_INCORRECT` (not sure but probably), so let me proceed #296 before taking action on this. Thanks suggestion!

tshelburne: Is this something that can be revisited? As more versions are published, it gets harder to keep up-to-date using a fork in a git reference.

yuhr:
Covered by #415.

PR #293, opened 2021-11-28
Enable coverage report

yuhr:


yuhr:
Seems like it's working without modification.

Issue #260, opened 2021-05-24
Better info to explain where object doesn't match the RunType


NickR49:
I have quite detailed RunType records for api responses. Unfortunately when something fails `record.check(values)` the exception message doesn't point out where the mismatch has occurred as follows -

```ValidationError: Expected { id: number; accessProfiles: { id: number; title: string; }[]; canEdit: boolean; clauseDetails: { id: number; documentId: number; documentTitle: string; documentType: string; documentStatus: string; documentLapseDate: string | null; documentExpiryDate: string | null; isCompliant: boolean; isRelevant?: boolean; isBreached: boolean; }[]; clauses: number[]; collections?: unknown[]; compliancePercentage: number; description: string | null; documentDetails: { id: number; type: string; title: string; status: string; lapseDate: string | null; expiryDate: string | null; clauseDetails: { id: number; isCompliant: boolean; isRelevant: boolean; isBreached: boolean; }[] | undefined; }[]; documents: number[]; filters?: { clause_ids: number[] | null; document_ids: number[] | null; document_statuses_f: string | null; document_types_f: string | null; due_from: null; due_relative_direction: ("+" | "-") | null; due_relative_duration: number | null; due_relative_start: number | null; due_to: null; expire_from: null; expire_to: null; favourited: boolean | null; ids: null; is_active: boolean | null; is_breached: boolean | null; is_completed: boolean | null; is_noncompliant: boolean | null; is_resolved: boolean | null; lapse_from: null; lapse_to: null; no_filter: boolean | null; owner_ids: number[] | null; watched: boolean | null; }; incidents: { id: number; documentId: number; documentTitle: string; documentType: string; documentStatus: string; documentLapseDate: string | null; documentExpiryDate: string | null; clauseId: number; }[]; isFavourited: boolean; isWatched: boolean; nbActiveDocuments: number; nbAllClauses: number; nbClausesWithIncidents: number; nbNonCompliantClauses: number; nbNonResolvedIncidents: number; owner: number; pins: { latitude: number; longitude: number; }[]; search?: null; tags: number[]; tasks: { id: number; documentId: number; documentTitle: string; documentType: string; documentStatus: string; documentLapseDate: string | null; documentExpiryDate: string | null; clauseId: number | null; incidentId: number | null; }[]; title: string; }, but was incompatible```

Nothing in the exception message narrows down the location of the mismatch - it just says `but was incompatible`.
Going through by hand trying to identify where a 65k JSON response doesn't match the record is painful.
[...]

PR #296, opened 2022-01-19
Verbose validation error message

HzrdIRL:
This PR addresses https://github.com/pelotom/runtypes/issues/260 - validation error messages should surface the incorrect fields without requiring extra error handling. 

Previously:
`Expected [number, { name: string; }], but was incompatible`

Now:
```
Validation failed:
[
  {
    "name": "Expected string, but was number"
  }
].
Object should match [number, { name: string; }]
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "coveralls": "^3.1.0",
     "jest": "26.6.3",
     "jest-github-actions-reporter": "^1.0.3",
+    "outdent": "^0.8.0",
     "prettier": "^2.2.1",
     "ts-jest": "^26.5.4",
     "typescript": "4.5.2"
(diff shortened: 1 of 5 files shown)


d33vil:
Any traction on this PR?

HzrdIRL:
> 

I would also like to know if this will get merged 😞

OmgImAlexis:
I've been using this personally because of how bad the error is.

```ts
import { ValidationError } from 'runtypes';
import type { RuntypeBase } from 'runtypes/lib/runtype';

export const validateRunType = <T = unknown>(runType: RuntypeBase, data: T): T => {
	try {
		return runType.check(data) as T;
	} catch (error: unknown) {
		if ((error as ValidationError).details) throw new Error(JSON.stringify((error as ValidationError).details));
		throw error;
	}
};
```

yuhr:
I'm very sorry for being unresponsible for a long time. Now I've got some time to work here.

PR #331, opened 2023-09-10
Feature/improved failure handling experience

zeeeeby:
Hello!

I have 2 ideas (and ready implementations in this PR) to improve error message readability.

## The first idea
I often encounter is overly detailed logging of complex structures in the "Expected" messages.

For example: 
let's have this schema
```
const Obj = Record({
  id: Number,
  name: String,
  nested: Record({
    id2: Number,
    name2: String,
    nested2: Array(
      Record({
        nested3: Record({
          id3: String,
        }),
      }),
    ),
  }),
});

// and we will validate this object
const result = Obj.validate({
		id: 3,
		name: "test" 
})
```
The validation result will be very messy:
```
{
      nested: 'Expected { id2: number; name2: string; nested2: { nested3: { id3: string; }; }[]; }, but was missing'
}
```

It seems unnecessary to describe the expected type since it is already known in the runtypes schema description. Moreover, this description complicates error message readability.

Instead, I propose writing "Expected (array | tuple | object | dictionary) but was ..."

in this PR result will change to this
```
{
      nested: 'Expected object, but was missing'
}
```
It looks readable and doesn't lose its meaning.

## The second idea
In messages like "Expected something but was (array | object)" it would be helpful to see exactly which entity was received.

I have decided to add a new field called `extraDetails` next to the `details` field, which will contain not only the error message but also the actual value (`received` field) and possibly the error code.

Let's go back to the previous schema and pass it this object as input:
```
const result = Obj.validate({
      id: 3,
      name: 5,
      nested: { nested2: { a: 'b' } },
  })
```
In this PR, the `result.extraDetails` field will contain this convenient message:
```
 {
      name: {
        message: 'Expected string, but was number',
        received: 5,
        code: 'TYPE_INCORRECT'
      },
      nested: {
        id2: {
          message: 'Expected number, but was missing',
          received: undefined,
          code: 'PROPERTY_MISSING'
        },
        name2: {
          message: 'Expected string, but was missing',
          received: undefined,
          code: 'PROPERTY_MISSING'
        },
        nested2: {
          message: 'Expected array, but was object',
          received: { a: 'b' },
          code: 'TYPE_INCORRECT'
        }
      }
  }
```
Thus, `extraDetails` will improve error perception without affecting the existing `details` field.

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -77,7 +77,7 @@ export type Failure = {
   details?: Details;
 
   /**
-   * The same as `details`, but with received value
+   * The same as `details`, but with more detailed error explanation
    */
   extraDetails?: Details;
 };
--- a/src/result.ts
+++ b/src/result.ts
@@ -67,9 +67,19 @@ export type Failure = {
   message: string;
 
   /**
+   * Received value
+   */
+  received: any;
+
+  /**
    * A detailed object enumerating where the validation failed exactly.
    */
   details?: Details;
+
+  /**
+   * The same as `details`, but with received value
+   */
+  extraDetails?: Details;
 };
 
 /**
--- a/src/runtype.ts
+++ b/src/runtype.ts
@@ -147,7 +147,7 @@ export function create<A extends RuntypeBase>(
   A.withGuard = withGuard;
   A.withBrand = withBrand;
   A.reflect = A;
-  A.toString = () => `Runtype<${show(A)}>`;
+  A.toString = () => `Runtype<${show(A, true)}>`;
 
   return A;
 
--- a/src/show.spec.ts
+++ b/src/show.spec.ts
@@ -150,7 +150,7 @@ const cases: [Reflect, string][] = [
 ];
 
 for (const [T, expected] of cases) {
-  const s = show(T);
+  const s = show(T, false);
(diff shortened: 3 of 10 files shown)


yuhr:
Thanks suggestion! Per-property failcode is one of missing thing in runtypes at this moment, so sounds good overall, but in the second idea, it sounds quite frightening to me to expose the received value into the error, because I'm pretty sure not a few users would be logging errors as is and sending the logs into somewhere insecure. Thus, at least `received` property should be explicitly enabled by a flag for each specific runtype the user intends to behave like that, such as `runtype.withOptions({ includeReceivedValueInFailure: true })`.

I'd say this should be worth beaking-changing the content of `details` property, for simplicity of the API surface. Besides, I'm now quite nervous and pessimistic at versioning i.e. I think every single change to any behavior that can be observable by users should be treated as a breaking change (even with [notice](https://github.com/pelotom/runtypes#error-information) not to depend on). So, if possible, I'd like to include this PR into the next major release.

yuhr:
The second idea has been implemented by #417 in a valid manner. I'd be seeking for even better idea to solve the verboseness of top-level messages.

Issue #229, opened 2021-04-19
Use SameValueZero equality


yuhr:
Currently runtypes use strict equality check i.e. `===` overall, but this way we cannot validate a value against `NaN` by using `Literal(NaN)` for example. It's better to use [SameValueZero](https://262.ecma-international.org/11.0/#sec-samevaluezero) equality which has more sane result. There's no dedicated API to perform it in JavaScript, but we can utilize [`Array.prototype.includes`](https://262.ecma-international.org/11.0/#sec-array.prototype.includes):

```ts
function sameValueZero(left, right) {
  return [left].includes(right);
}
```

SIGUSR97:
Consider using the [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) function.

PR #392, opened 2025-01-02
Fix Literal TypeError on null prototype objects

72636c:
Works around the following issue:

```typescript
String(Object.create(null))
// Uncaught TypeError: Cannot convert object to primitive value

Literal(1).validate(Object.create(null))
// Uncaught TypeError: Cannot convert object to primitive value
```

We encountered this issue when working with [Apollo Server](https://github.com/apollographql/apollo-server). It appears to create arguments with a null prototype 🤷.

I've cleaned up a comment regarding `Symbol()` stringification that does not apply to the `String()` constructor as per added test case and MDN.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/String#using_string_to_stringify_a_symbol

### DIFF
--- a/src/Literal.ts
+++ b/src/Literal.ts
@@ -16,8 +16,8 @@ interface Literal<T extends LiteralBase = LiteralBase> extends Runtype.Common<T>
 const literal = (value: unknown): string =>
 	Array.isArray(value)
 		? globalThis.String(value.map(literal))
-		: value !== null && typeof value === 'object' && Object.getPrototypeOf(value) === null
-			? '[object null]'
+		: value !== null && typeof value === "object"
+			? "[object Object]"
 			: typeof value === "bigint"
 				? globalThis.String(value) + "n"
 				: globalThis.String(value)
--- a/src/Literal.ts
+++ b/src/Literal.ts
@@ -13,15 +13,14 @@ interface Literal<T extends LiteralBase = LiteralBase> extends Runtype.Common<T>
 	value: T
 }
 
-/**
- * Be aware of an Array of Symbols `[Symbol()]` which would throw "TypeError: Cannot convert a Symbol value to a string"
- */
-const literal = (value: unknown) =>
+const literal = (value: unknown): string =>
 	Array.isArray(value)
-		? globalThis.String(value.map(globalThis.String))
-		: typeof value === "bigint"
-			? globalThis.String(value) + "n"
-			: globalThis.String(value)
+		? globalThis.String(value.map(literal))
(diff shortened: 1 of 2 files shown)


src/Literal.ts
```diff
@@ -13,15 +13,14 @@ interface Literal<T extends LiteralBase = LiteralBase> extends Runtype.Common<T>
...
- * Be aware of an Array of Symbols `[Symbol()]` which would throw "TypeError: Cannot convert a Symbol value to a string"
- */
-const literal = (value: unknown) =>
+const literal = (value: unknown): string =>
 	Array.isArray(value)
-		? globalThis.String(value.map(globalThis.String))
-		: typeof value === "bigint"
-			? globalThis.String(value) + "n"
-			: globalThis.String(value)
+		? globalThis.String(value.map(literal))
```
72636c, line 18: Recursing `literal` will handle null prototype objects, BigInt formatting, etc. within an array.

Are there any potential issues here? I'm just cautious because it wasn't implemented this way before 🤔

src/Literal.ts
```diff
@@ -13,15 +13,14 @@ interface Literal<T extends LiteralBase = LiteralBase> extends Runtype.Common<T>
...
+		: value !== null && typeof value === "object"
+			? "[object Object]"
```
72636c, line 20: This short circuits before getting to the `String(value)` that throws an error. I was initially thinking of checking `Object.getPrototypeOf(value) === null` but that may not handle exceptional scenarios that [extend null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/extends#extending_null).

src/Literal.ts
72636c: Aha, CI says yes 🙂

72636c: I can think of two approaches here:

1. Just `catch` the error when it crops up and return some placeholder string. Hopefully still be preferable over throwing a `TypeError`.
2. Stop unpacking a nested array after a certain depth. `literal` currently flattens the array so I'm not sure how valuable its output actually is in practice.

yuhr:
Thank you for adding test cases, this is very helpful.

yuhr:
For now, I'd keep this PR small enough to handle null prototype objects. Further overhaul would be done in another PR.

yuhr:
Just to note, the equality used in validation will be changed to SameValueZero in v7, as planned in #229. So `Literal(NaN).validate(NaN)` will become passing successfully.

72636c:
Amazing, thank you @yuhr!

PR #399, opened 2025-01-04
refactor: remove unnecessary `any`

yuhr:


yuhr:
Hm, noticed it's actually necessary 😅

PR #432, opened 2025-01-10
chore: use unique symbol for `ValidationError` temporarily

yuhr:


yuhr:
Use of non-unique symbols is postponed until <https://github.com/microsoft/TypeScript/issues/60944> is resolved.

